Issue #3, opened 2016-04-08
Redux/redux-thunk must be npm-installed on the host app


brettburley:
@toranb thanks for putting this together, this is really awesome!

From my experience with using this addon, I had to `npm install` `redux` and `redux-thunk` for it to compile properly.

It appears that this is a known limitation with `ember-browserify`:
https://github.com/ef4/ember-browserify
> To have the host app consume the npm dependency, ember-browserify and the npm dependency must be installed in the host app as well AND the module must be imported from within the app/ directory of the addon. So in your app, simply npm install ember-browserify and whatever npm dependencies you need consumed by ember-browserify.

It doesn't look like https://github.com/toranb/ember-redux-example/ includes `redux` or `redux-thunk` in it's `package.json`, so I'm curious about your experience with this.

toranb:
@brettburley honestly I assumed having `redux-thunk` in the [dependencies][] block of the package.json was enough. When I install this in any modern ember app I don't get any type of compile error with ember-cli.

What versions of ember-cli / ember / and ember-browserify are you using? If you use the versions you see below everything should be solid? Here are 2 examples worth looking at

ember-cli v2.4.2 and v2.4.3
ember-browserify v1.0.2
ember v2.4

https://github.com/toranb/ember-redux-ddau-example/blob/master/package.json
https://github.com/toranb/ember-redux-example/blob/master/package.json

[dependencies]: https://github.com/toranb/ember-redux/blob/master/package.json#L57

brettburley:
Both of those repos failed for me locally (Max OS X). It appears to be a result of the latest `ember-browserify v1.1.8`. If I downgrade to `v1.0.2` as suggested, it seems to work fine.

For reference, here's the error I get with `v1.1.8`:
```
The Broccoli Plugin: [object Object] failed with:
Error: Cannot find module 'redux' from '/Users/brettburley/Git/ember-redux-ddau-example/tmp/core_object-input_staging-zXLMsE6R.tmp'
```

I haven't investigated to see why `ember-browserify` seems to have regressed (although the behavior I'm seeing with `v1.1.8` seems to be consistent with their documentation).

toranb:
Huh so I just bumped both of those examples to ember-browserify v1.1.8 and they work w/out error

https://github.com/toranb/ember-redux-ddau-example/commit/d9f87e3882576c4ea4f72f5e8f57493563651947

[...]

PR #4, opened 2016-04-08
[BUG] Add support for Thunk 2.x series.

brettburley:
Thunk 2.x now requires looking at the default key when using CommonJS
imports (which is used by ember-browserify):
https://github.com/gaearon/redux-thunk/releases/tag/v2.0.0

When using Thunk 2.0.1, this is the error I saw as a result:
```
applyMiddleware.js:48 Uncaught TypeError: middleware is not a function
```

### DIFF
--- a/app/middleware/index.js
+++ b/app/middleware/index.js
@@ -1,3 +1,6 @@
 import thunk from 'npm:redux-thunk';
 
+// Thunk 2.x exports on a default key.
+const resolvedThunk = thunk.__esModule ? thunk.default : thunk;
+
 export default [thunk];


toranb:
I've looked at upgrading this but because we bundle redux-thunk with this addon I assumed we would upgrade this along w/ the middleware itself. 

Along those lines - what exactly is the reason to upgrade from 1.0.3 to 2.x ?

https://github.com/gaearon/redux-thunk/commits/master

brettburley:
Ah good point. I think this is a result of me having to install `redux-thunk` on my app itself (which then doesn't use the 1.x series as specified by the package.json here) due to https://github.com/toranb/ember-redux/issues/3. Let me take a look at that issue to better understand it.

toranb:
@brettburley also about "why upgrade to `redux-thunk` v2" - I asked a question to better understand the value in this upgrade

https://github.com/gaearon/redux-thunk/issues/67

toranb:
I looked into this further and fixed the issue you brought up in the commit below

https://github.com/toranb/ember-redux/commit/2a70e1481b6759e1a88fcbea9adbbd7f3f72d55a

With this in place both redux-thunk 1.0.3 and 2.0.1 work as you would expect. I also updated the README to show `redux-thunk` v2.0.1 as the default going forward

toranb:
This is now available in `ember-redux` v1.2.0

https://github.com/toranb/ember-redux/commit/9911093ab2ac96da714aa18cd5dd8bcce2bcd89f

Issue #5, opened 2016-04-08
Return dispatch return value from service?


brettburley:
When using the `redux-thunk` middleware, I've created some action creators with routes that look like:
```javascript
import Ember from 'ember';
import route from 'ember-redux/route';

function loadData(id) {
  return async function (dispatch) {
    const data = await fetch(...);
    dispatch({ type: 'DESERIALIZE_DATA', data: data });
  }
}

export default route({
  model(dispatch, params) {
    return dispatch(loadData(params.id));
  }
})(Ember.Route.extend());
```

However, since the dispatch method on the service (https://github.com/toranb/ember-redux/blob/master/app/services/redux.js#L20) just returns undefined, the ember model loading hooks don't wait for the async action to complete before rendering the route.

Are there any downsides to modifying `dispatch` on the service to always return the dispatch result? The only consideration I can think of right now is potentially returning a value in the model hook, which would cause Ember to setup a model on the controller. However, so long that you don't `return` dispatch from the model *unless* it's a promise from thunk, it seems like we can avoid inadvertently returning a model.

toranb:
This is a good catch - I just realized I don't have any examples/tests that use this pattern in the addon. To add this I'd just like a legit example (codified as a test case w/ acceptance) just so others who follow can see it in action.

In the above I see you are using fetch but I'm not sure what ember addon/npm module you are using. Would you be up to add a component/route/acceptance test here that shows it off? 

One other question about your example - in the `loadData` function I see you return `async` function but inside that function you simply `dispatch` but does this need to be `return dispatch` instead?

If you can create a full example app (instead of the acceptance test) I could take that working example and test drive this feature. Just let me know how you prefer to operate going forward

brettburley:
Ah yes, I was just using `fetch` as an example of something that returns a promise:
https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API

Regarding calling `dispatch` vs. returning it, I think I *don't* actually want to return it, because if I do, I would expect Ember will use that resolved promise value as the `model`. By not returning it, the promise resolves with `undefined`, and therefore we don't get any unnecessary `model` attached to the controller.

brettburley:
Closing this in favor of PR https://github.com/toranb/ember-redux/issues/7.

PR #7, opened 2016-04-09
[FEATURE]: return dispatch value to allow model hook to wait for thunk promises

brettburley:
When calling an action that returns a promise (ala redux-thunk), we must return that promise from the dispatch action so that we can use it in a model hook to take advantage of the Ember model loading hooks.

https://github.com/toranb/ember-redux/issues/5


toranb:
This looks great! The only changes Id suggest before we pull this in

1) move the dynamic route out of the test and into the dummy app as a rea file 
2) remove the actions (loading) and use a real hbs each loop to confirm the users.all was setup correctly
3) add a link-to in the application.hbs file to /dynamic

I only considered asking for this change to help others who follow see a more consistent pattern of using the dummy app. I realize this isn't for everyone but I like having the ability to point at this addon and tell people who are less familiar with testing "just use ember s"

brettburley:
Moved this route into the actual dummy app.

I did, however, keep the `loading` action in the route. This is invoked by Ember automatically when the route goes into a loading state, and I think it's important to catch this in this particular case. This is the part of the test that fails when the service's `dispatch` method *doesn't* return the promise. So by verifying that the loading state is invoked, we can be sure that we're properly returning a promise by the `dispatch` in the route's `model` hook.

I was unable to find a better way to determine if the route goes into a loading state... it's tough to catch in the acceptance test itself since any `andThen`s will wait for all promises to resolve (and therefore missing the loading substate completely).

toranb:
@brettburley ah that makes total sense! Sorry for the lack of understanding on my part earlier!

This pull request looks solid my friend! Thanks for the extra work involved to educate people like me who are not as familiar w/ action creators yet :)

toranb:
I'll merge this / bump the version, tag and publish it to npm tonight! Thanks again!

toranb:
I just published this as `v1.3.0` on npm :) thanks again for all the hard work that went into this pull request!

Issue #12, opened 2016-05-08
Access to ember's container for Action creators


opsb:
My app makes use of phoenix channels (over a websocket) and needs to manage subscribing/dispatching to them. Typically with redux you'd use Action Creators to manage async operations, dispatching actions to the reducer at the appropriate times. These Action Creators are shared between components to allow the use of common business logic / resources. 

With ember-redux `dispatchToActions` fills the role of the Action Creators. Because it's just a function there's nothing to stop you importing a map of ActionCreators that's shared across the entire app. One limitation though is that it doesn't have access to ember's container. Ember's container seems like the natural place to keep track of things like websockets, channel subscriptions etc. so this would be a useful addition. 

I see a few options here:
1) Simply expose the ember container with `dispatchToActions(dispatch, container)`
2) Move the dispatcher into the ember container and use a convention for registering actions e.g.

```
// /app/actions/todos
export default Ember.Actions.extend({
  channels: inject.service('channels'),
  dispatch: inject.service('dispatch'),

  addTodo(name) {
    let {dispatch, channels} = this.getProperties('dispatch', 'channels');
    let todoId = uuid();
    dispatch({type: 'ADD_TODO_REQUESTED', todoId});
    channels.get('todos').add(name)
      .receive('ok', resp => dispatch({type: 'ADD_TODO_COMPLETED', todoId}))
      .receive('error', resp => dispatch({type: 'ADD_TODO_FAILED', todoId})
  }
});
```

(2) could potentially just be built on top of (1) so perhaps that would be a good place to start.

There's most likely other approaches here and I'd be very happy to discuss them. If we can come to a consensus then I should have time to put a PR together this week.

opsb:
It occurred to me that I could just use standard ember services to handle action creation (rather than redux-thunk) e.g.

```
// app/services/todos.js
export default Ember.Service.extend({
  redux: inject.service('redux'),
  channels: inject.service('channels'),

  add(name) {
    let {redux, channels} = this.getProperties('redux', 'channels');
    let todoId = uuid();
    redux.dispatch({type: 'ADD_TODO_REQUESTED', todoId});
    channels.get('todos').add(name)
      .receive('ok', resp => redux.dispatch({type: 'ADD_TODO_COMPLETED', todoId}))
      .receive('error', resp => redux.dispatch({type: 'ADD_TODO_FAILED', todoId});
  }
});
```

This service can then be injected into a component meaning the component would be completely ignorant of redux. 

[...]

PR #13, opened 2016-05-15
Cookbook and ember services

opsb:
Docs for using ember services, as requested in https://github.com/toranb/ember-redux/issues/12


toranb:
I merged this in and updated it to have the consistent JS syntax highlighting - thanks for submitting this awesome pattern :)

PR #14, opened 2016-07-20
Activate redux-devtools-extension (if available)

fpauser:


toranb:
Hey thanks for the pull request! It looks like I've done a poor job documenting the enhancers extension point inside ember-redux that lets you opt in for dev tooling.

Here is an example of it

https://github.com/toranb/ember-redux-ddau-example/blob/master/app/enhancers/index.js

I prefer this to adding more complexity inside the redux service itself. That said, I think you've highlighted the missing pull request to the docs GH-pages site on "time travel/dev tooling"

You mind opening a pull request for the documentation so others see the idiomatic enhancer option mentioned above?

Sorry for the lack of docs about dev tooling this far!

fpauser:
I just overlooked http://www.ember-redux.com/configure/, so from my standpoint this is clear enough and needs no further doku.

PR #15, opened 2016-08-04
Create default blueprint

dustinfarris:
This enables a one-line install of ember-redux

TEST WITH:

```
ember install "ember-redux@dustinfarris/ember-redux#c2d0055"
```

CHANGES:

- Create default blueprint which adds dependencies to user's own package.json
- Move internal dependencies to devDependencies

### DIFF
--- a/blueprints/.jshintrc
+++ b/blueprints/.jshintrc
@@ -1,0 +1,6 @@
+{
+  "predef": [
+    "console"
+  ],
+  "strict": false
+}
--- a/blueprints/ember-redux/index.js
+++ b/blueprints/ember-redux/index.js
@@ -1,0 +1,18 @@
+/*jshint node:true*/
+module.exports = {
+  description: 'Installation blueprint for ember-redux',
+  normalizeEntityName: function() {},
+
+  afterInstall: function() {
+    return this.addAddonsToProject({
+      packages: [
+        {name: 'ember-browserify', target: '^1.1.11'},
+      ]
+    }).then(function() {
+      return this.addPackagesToProject([
+        {name: 'redux', target: '^3.5.2'},
+        {name: 'redux-thunk', target: '^2.1.0'}
+      ]);
+    }.bind(this));
+  }
+};
--- a/README.md
+++ b/README.md
@@ -9,10 +9,7 @@
 ## Installation
 
 ```
-npm install ember-browserify -D
-npm install redux@3 -D
-npm install redux-thunk@2 -D
-npm install ember-redux -D
+ember install ember-redux
 ```
 
 ## Documentation
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "ember-resolver": "^2.0.3",
     "loader.js": "^4.0.0",
     "broccoli-asset-rev": "^2.2.0",
+    "ember-browserify": "^1.1.11",
     "ember-cli-app-version": "^1.0.0",
     "ember-cli-dependency-checker": "^1.2.0",
     "ember-cli-htmlbars": "^1.0.1",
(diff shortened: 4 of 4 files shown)


toranb:
this is awesome !!! thank you 100x sir

toranb:
I'm planning to include this and another breaking change I found today as the v2.0.0 and I'm planning to ship that tonight if all goes well :)

dustinfarris:
Awesome.  I think once you do that this should work with ember twiddle also.

blueprints/ember-redux/index.js
```diff
@@ -0,0 +1,18 @@
...
+  normalizeEntityName: function() {},
+
+  afterInstall: function() {
+    return this.addAddonsToProject({
+      packages: [
+        {name: 'ember-browserify', target: '~1.1.11'},
+      ]
+    }).then(function() {
+      return this.addPackagesToProject([
+        {name: 'redux', target: '~3.5.2'},
```
toranb: any reason we are using the `~` instead of `^` here ?

blueprints/.jshintrc
```diff
@@ -0,0 +1,6 @@
+{
+  "predef": [
```
toranb: what does this little guy do exactly? just curious

toranb:
The only downside I see right now is that it's adding a very set version to the package.json for the consuming app. Any way we can have it add a `^` in front after the ember install finishes?

here was my diff after running it (I like the version bumps but prefer if they pull in latest using `^`)

screen shot 2016-08-04 at 7 44 43 pm

blueprints/ember-redux/index.js
dustinfarris: No reason.  I can change it if you like

blueprints/.jshintrc
dustinfarris: this file is added by ember-cli as a result of:

```
ember generate blueprint ember-redux
```

toranb: ah, ok

blueprints/ember-redux/index.js
toranb: cool - does this solve the other issue I mentioned in the main thread?

dustinfarris:
@toranb Changing the blueprint target versions to use a `^` affects which version is written to the user's package.json, but does not actually add the range selector (for better or worse).

e.g. if we have `{name: "redux", target: "~3.0.0"}` in the blueprint, `"redux": "3.0.6"` will be written to the user's package.json.  if we have `{name: "redux", target: "^3.0.0"}`, in the blueprint, `"redux": "3.5.2"` will be written.

blueprints/ember-redux/index.js
dustinfarris: No.  Please see my response and confirm if you want this change or not.

toranb: ah, well I think this is still a HUGE improvement to make the addon more beginner friendly. Lets merge it anyway :) and because it's not breaking I'll just bump to v1.5.0

sound good?

toranb:
@dustinfarris you mind pushing up a commit to do the `^` change mentioned and then I'll merge this? thanks again for all the work you put into this - seriously!

blueprints/ember-redux/index.js
toranb: Oh i just realized why we had that show up now (likely) - your first
reducer wasn't correct (remember we had to swap user.js and index.js in
that file because they were reversed?) That would be enough to confused
redux/ likely having that warning show up (if not an error like I'd expect)

Toran

dustinfarris:
@toranb updated and rebased.  should be good to go!

toranb:
@dustinfarris awesome! I'll merge it after we go green in a few sec

Issue #17, opened 2016-08-19
Consider exploding args in the route.js function


foxnewsnetwork:
Background
===
In your code at https://github.com/toranb/ember-redux/blob/master/addon/route.js#L12 we see:

```javascript
route[func] = function(args) {
  return funcs[func](redux.dispatch.bind(redux), args);
};
```

Notably, the `args` param only captures the first parameter of whatever native `Ember.Route` method we're declaring. Consider making the following changes:

Suggested Changes
===
```javascript
route[func] = function(...args) {
  args.unshift(redux.dispatch.bind(redux));
  return funcs[func].apply(route, args);
};
```

Reasoning
===
`Ember.Route` methods such as `beforeModel`, `afterModel`, etc., all get passed not only the model, but also the transition, and other such possible parameters. Properly passing those parameters in may be useful for users.

Questions
===
Q: What do you think? If you're okay with it, I'll can do a PR when I get off of work

Q: Why not also pass in `redux.subscribe`? Is using `subscribe` in the route not recommend?

Notes
===
Thanks for your guidance in the slack chat the other day, I looked into redux-saga and it seems more and more that's the way we'll wind up going with.

toranb:
@foxnewsnetwork wow! I truly assumed that I was actually passing all the arguments through but as you pointed out ... clearly not :)

I view this behavior as a bug honestly - it was not my intention to alter the contract or arguments of a route in any way (as ember developers should expect everything they would get normally w/ the addition of dispatch).

Regarding subscribe - I'd like to leave that out to ensure we only have 1 place for the developer to register for callbacks. The connect api for components is the place to listen for changes/ re-render. In ember I've always felt the role of the router was simply to fetch data given some url

I'd welcome a pull request if you've got time to show it in action. You'll notice I usually prefer a higher level test but if you find that difficult to pull off just write something at the unit level (I was able to write a failing test myself just now using setupController because the 3rd argument is undefined (not the model as you'd expect) so anything you'd normally do in that method using classic ember would fall down quickly as you might imagine)

Thanks for hearing me out on slack and deciding to take the addon for a spin :)

toranb:
Thanks again - your PR getting merged now closes this out :)

PR #18, opened 2016-08-19
17 - fixed arguments like transition not being passed into the route hook

foxnewsnetwork:
Reference
===
https://github.com/toranb/ember-redux/issues/17#issuecomment-240909216

Changes
===
* properly exploded input arguments and now using apply to call them in `addon/route.js`
* added an acceptance-test for route-hooks
* modified the item-list component and items route to accomodate for the afterModel test
* added a reducer to handle `AFTER_MODEL`

Unrelated Note
===
I was looking at the `addon/components/connect.js:41` when I came across this:

```javascript
                    this.unsubscribe = redux.subscribe(() => {
                        props.forEach(function(name) {
                            component.notifyPropertyChange(name);
                        });
                    });
``` 
And it looks like each time an action is sent to redux, we tell ember that the property changed even if the property didn't actually change... is this ok? I mean, granted I'm making a lot of assumptions without any first-hand experience, but presumably, in a big app where actions are fired all the time, ember's going to have to be really busy all the time with updating all the computed properties everywhere on every action. Do you ever have any troubles with this in your applications?

### DIFF
--- a/addon/route.js
+++ b/addon/route.js
@@ -9,8 +9,9 @@ var route = function(funcs) {
                 var redux = this.get('redux');
                 var route = this;
                 Object.keys(funcs).forEach(function(func) {
-                    route[func] = function(args) {
-                        return funcs[func](redux.dispatch.bind(redux), args);
+                    route[func] = function(...args) {
+                        args.unshift(redux.dispatch.bind(redux));
+                        return funcs[func].apply(route, args);
                     };
                 });
                 this._super(...arguments);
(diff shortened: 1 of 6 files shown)


toranb:
@foxnewsnetwork this is great work and again I really really appreciate you taking the time to find this bug/ fix it and write tests to show it in action :)

Regarding the performance question - that's very much on my mind and I know react-redux is doing [something](https://github.com/reactjs/react-redux/blob/master/src/components/connect.js#L250) to prevent a full on virtualdom thrashing so I have no doubt we will land something in the next month or 2 that solves this. Right now I truly don't see any performance issues but I'm not exactly building something of scale yet as you might imagine

toranb:
This was published and should be available as of v1.5.2

foxnewsnetwork:
Awesome, thanks dude.

toranb:
@foxnewsnetwork just landed the subscribe changes you mentioned here in #23

PR #20, opened 2016-08-29
Add reducer blueprints

jkusa:
This adds blueprints to help manage reducers.

Create `app/reducers/index.js` on addon install:
```
➜  ember-redux-todomvc git:(master) ✗ ember generate ember-redux default
installing ember-redux
  create app/reducers/index.js  
```

Create a new reducer, its test, and update `app/reducers/index.js` :
```
➜  ember-redux-todomvc git:(master) ✗ ember generate reducer count
installing reducer
  create app/reducers/count.js
updating app/reducers/index.js
  add count
installing reducer-test
  create tests/unit/reducers/count-test.js
```

Remove a reducer, its test, and update `app/reducers/index.js`:
```
➜  ember-redux-todomvc git:(master) ✗ ember destroy reducer count
uninstalling reducer
  remove app/reducers/count.js
updating app/reducers/index.js
  remove count
uninstalling reducer-test
  remove tests/unit/reducers/count-test.js
```

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -14,3 +14,4 @@
 bower.json
 ember-cli-build.js
 testem.js
+node-tests/
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,3 +37,4 @@ install:
 
 script:
   - ember try $EMBER_TRY_SCENARIO test
+  - npm run-script test:blueprints
--- a/blueprints/ember-redux/files/__root__/reducers/index.js
+++ b/blueprints/ember-redux/files/__root__/reducers/index.js
@@ -1,0 +1,1 @@
+export default {};
--- a/blueprints/reducer-test/files/tests/unit/reducers/__test__.js
+++ b/blueprints/reducer-test/files/tests/unit/reducers/__test__.js
@@ -1,0 +1,12 @@
+import reducer from '<%= dasherizedModulePrefix %>/reducers/<%= dasherizedModuleName %>';
+import { module, test } from 'qunit';
+
+module('<%= friendlyTestName %>');
+
+test('initial state', function(assert) {
+  assert.expect(1);
+
+  assert.deepEqual(reducer(undefined, {}),
+    {}, //replace with actual value
+  '<%= camelizedModuleName %> reducer returns the correct initial state');
+});
--- a/blueprints/reducer-test/index.js
+++ b/blueprints/reducer-test/index.js
@@ -1,0 +1,15 @@
+/*jshint node:true*/
+
+var stringUtils = require('ember-cli-string-utils');
+var testInfo    = require('ember-cli-test-info');
+
+module.exports = {
+  description: 'Generates a reducer test',
+
+  locals: function(options) {
+    return {
+      friendlyTestName: testInfo.name(options.entity.name, 'Unit', 'Reducer'),
+      dasherizedModulePrefix: stringUtils.dasherize(options.project.config().modulePrefix)
+    };
+  }
+};
--- a/blueprints/reducer/files/app/reducers/__name__.js
+++ b/blueprints/reducer/files/app/reducers/__name__.js
@@ -1,0 +1,9 @@
+const initialState = {}; //replace with actual value
+
+export default ((state=initialState, action) => { // jshint ignore:line
+  switch (action.type) {
+    //handle actions
+    default:
+      return state;
+  }
+});
--- a/blueprints/reducer/index.js
+++ b/blueprints/reducer/index.js
@@ -1,0 +1,143 @@
+/*jshint node:true*/
+
+var fs       = require('fs');
+var EOL      = require('os').EOL;
+var recast   = require('recast');
+var builders = require('recast').types.builders;
+var chalk    = require('chalk');
+
+module.exports = {
+  description: 'Generates a reducer',
+
+  reducerIndexPath: 'app/reducers/index.js',
+
+  updateReducerIndex: function(updateAst) {
+    var file;
+
+    try {
+      file = fs.readFileSync(this.reducerIndexPath, 'utf8');
+    } catch (err) {
+      if (err.code !== 'ENOENT') throw err;
+
+      var error = new Error( 'Cannot find ' + this.reducerIndexPath +
+        ' Try running: ember generate ember-redux default');
+
+      this.ui.writeError(error);
+      return;
+    }
+
+    this.ui.writeLine('updating ' + this.reducerIndexPath);
+    var ast  = recast.parse(file);
+    updateAst(ast);
+
+    var code = recast.print(ast).code;
+    fs.writeFileSync(this.reducerIndexPath, code);
+  },
+
+  afterInstall: function(options) {
+    var reducerName = options.entity.name;
+
+    this.updateReducerIndex(function(ast) {
+      insertImport(reducerName, ast);
+      insertExportProp(reducerName, ast);
+      this._writeStatusToUI(chalk.green, 'add', reducerName);
+    }.bind(this));
+  },
+
+  afterUninstall: function(options) {
+    var reducerName = options.entity.name;
+
+    this.updateReducerIndex(function(ast) {
+      removeImport(reducerName, ast);
+      removeExportProp(reducerName, ast);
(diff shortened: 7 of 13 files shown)


toranb:
@jkusa wow man this is an epic pull request! A few questions

1) I see you pulled in mocha but is it truly required?
2) instead of the usual jshint ignore example could we instead do something like this to avoid the linting noise?

```js
export default ((state, action) => {
    switch (action.type) {
    //handle actions
    default:
        return state || initialState;
    }
});
```

3) in the reducer test I see this line and was curious what it would look like after a merge

```
//TODO replace this step with default blueprint once issue fixed
```

4) why is this required/ what is it used for? `babel-plugin-transform-es2015-arrow-functions`

5) We went from just one dependency `"ember-cli-babel": "^5.1.5"` to 6 including a test dependency -is this absolutely required for the blueprint work? any cons to be aware of taking this on?

6) what documentation changes would/should go with this in the gh-pages branch to document how they are used/ etc ?

jkusa:
@toranb thanks for looking at this. It is somewhat larger than I originally expected.  :)

My hope is that this can assist ember devs in getting their feet wet with ember-redux!

1) This comes from the [ember-cli/ember-cli-blueprint-test-helpers](https://github.com/ember-cli/ember-cli-blueprint-test-helpers) blueprint which provides a framework for testing blueprints. This is what is used in the core ember repos and a few community ones as well. 

2) Sounds good to me.

3) I was hoping to simply run the default blueprint instead of manually creating the index file. This would look like:

```js
.then(() => emberGenerate((['ember-redux', 'default'])))
```

4) This is also generated by [ember-cli/ember-cli-blueprint-test-helpers](https://github.com/ember-cli/ember-cli-blueprint-test-helpers). Instead of pulling in all of babel, only some of the es2015 babel transforms are pulled in and executed. This actually shaves off 3-4 seconds when running the tests compared to using [babel-preset-es2015](http://babeljs.io/docs/plugins/preset-es2015/). Specifically, `babel-plugin-transform-es2015-arrow-functions` lets you defined functions with `=>`. We could use `babel-preset-es2015` or simply use es5 and not bring in babel. Thoughts?

5) Yeah, the increase in dependencies did cross my mind. Here is the breakdown:

| package                | functionality                                                                               |
| ---------------------- | ------------------------------------------------------------------------------------------- |
| recast                 | This is the lib to do the ast manipulation. This is what the update `router` blueprints use |
| chalk                  | This is used to color output text red & green just like the `router` blueprint                 |
| ember-cli-string-utils | This is used to dasherize the reducer name in the generated test                                 |
| ember-cli-test-info    | This is used to print the test module name in the same way other stock blueprints do         |

As you can see the core ember libs have similar dependencies:
https://github.com/emberjs/ember.js/blob/master/package.json#L54
https://github.com/emberjs/data/blob/master/package.json#L25

It is possible to remove all of the new dependencies other than `recast`. The other dependencies provide a consistent behavior when compared to other ember blueprints. 

As for cons, none come to mind other than there are now more dependencies. That said, I feel it shouldn't be an issue due to the fact that `ember-redux` is an ember addon and the dependencies added either come from the `ember-cli` org or are used by core ember repos.

6) As for documentation, I think the examples in the PR description should be somewhere. Currently, the gh-pages has the `introduction` section that explains redux generally and then the `quickstart` section that picks up from the intro and explains how the addon does some of the wiring. Would it make sense to have a tutorial that shows how to use this addon from scratch? The blueprints could be used to add reducers.

toranb:
@jkusa sorry for what seems like a slow reply to this pull request all week. To be completely honest I'm wrestling with the idea of merging it. On the one hand you've done a great job and communicated 100x - losing a contributor and community member with your skill set and desire would be harmful (assuming you would be hurt if I did not merge this). On the other hand when I ask "what problem does this solve? and what baggage comes with the solution?" I boil it all down to "well it seems to add a single reducer that beyond hello-world isn't super helpful for the new ember-redux user"

If I seem off base here/ or if I've somehow reduced your work let me know, but I often try to look at a general "problem statement" when I add a pull request of this size/ with additional dependencies. At the moment I'm stuck asking myself "do I want to maintain this addition/ is the juice worth the squeeze"?

dustinfarris:
I have to agree with @toranb's sentiment.  Definitely a kick-ass PR, but not sure if it fits for the long-term.

I do like the idea of the default blueprint creating `reducers/index.js`.

toranb:
@jkusa giving this some additional thought, I think @dustinfarris has made a good point. Why not provide a default reducer when you "ember install" ? This would be more helpful than what you experience today - not sure how you feel about that vs what you've already done in the pull request here...

jkusa:
@toranb sorry for taking so long to reply. I have been away from a computer for the last few days.

I realize this PR provides a relatively small convenience. It is truly a "nice to have". This is somewhat true for most blueprints. However, I do appreciate the convenience/automation they provide :)

I think one of Ember's strengths is the focus on tooling (including blueprints) and conventions across the community. Since this is an ember addon to assist with the integration with redux, and adding reducers as an app evolves can be typical, I felt this blueprint was a natural fit and an easy way to support this project. I also thought additional blueprints could be added for the other redux constructs in the future.

Having said that, I do understand your concerns. It is the management of the `index.js` that brings the bulk of the logic and need for testing and consequently the dependencies. Perhaps a convention could remove the need for maintaining an `index.js`.

I will not be upset if you do not merge this in. I never created a gh issue. I was originally going to add this to: jkusa/ember-redux-helpers which uses this addon, but then thought it was better suited here.

If you rather just have the default blueprint add a `reducers/index.js` that exports an empty object as done in this PR, I can do that in a new PR.

toranb:
@jkusa I think the default reducers/index.js is the ideal solution here. If it's cool with you, you mind closing this out and opening another that just adds a simple reducers/index.js file?

jkusa:
@toranb, sure. I can open a new PR with the default reducers/index.js found in this PR:

https://github.com/jkusa/ember-redux/blob/7392c1e6f98cf73e22a7fb4c70fa6fa680956688/blueprints/ember-redux/files/__root__/reducers/index.js

toranb:
@jkusa I think that solves the most basic issue. I do hear you about advanced usages (to generate middleware/ reducers/ enhancers) but they are so dead simple (mostly an export) - do people really use this/ need this when the resulting file is just a 1 liner exporting a vanilla JS object or function? </serious question here/ feeling out of touch at times>

jkusa:
> @jkusa I think that solves the most basic issue.

@toranb right. It creates the `app/reducers` directory and removes the error: 

```
Uncaught Error: Could not find module `ember-redux-test/reducers/index` imported from `ember-redux-test/services/redux`
```
when trying to use the `redux` service when `app/reducers/index.js` is not present.

Is this what you were thinking as a simple `reducers/index.js` (maybe with a small comment in the file explaining what `combineReducers` is expecting)? Or did you have something else in mind.

>I do hear you about advanced usages (to generate middleware/ reducers/ enhancers) but they are so dead simple (mostly an export) - do people really use this/ need this when the resulting file is just a 1 liner exporting a vanilla JS object or function?

I'm guessing you are referring to my earlier [comment](https://github.com/toranb/ember-redux/pull/20#issuecomment-245415959). As for is this used/needed. Required, no. Used, I guess it depends on the dev. I've been using Ember for a while and I still use the ember.js generators as a part of my work flow. They create the implementation and test files in the correct directories. Automate the wiring (ex: the router or the index.js in this PR). Help with file consistency. For new devs, they provide some guidance. 

In my experience, many of the blueprints/generators result in a fairly basic file:
 - [ember-component](https://github.com/emberjs/ember.js/blob/master/blueprints/component/files/__root__/__path__/__name__.js)
 - [ember-util](https://github.com/emberjs/ember.js/blob/master/blueprints/util/files/__root__/utils/__name__.js)
 - [ember-data-model-fragments](https://github.com/lytics/ember-data-model-fragments/blob/master/blueprints/fragment/files/app/__path__/__name__.js)
 - [mirage-fixture](https://github.com/samselikoff/ember-cli-mirage/blob/master/blueprints/mirage-fixture/files/__root__/fixtures/__name__.js)

In the case of redux the functions are simple, but they do have contracts:

```js
export default ((state=initialState, action) => { ... }
```

```js
export default store => next => action => { ... }
```

PR #22, opened 2016-09-14
reducer and state-initializer generators

ksnyde:
I haven't really tested this at all yet but I wanted to get this out there early to gauge your interest. I think it would be really nice if we could add all the standard templates one would need for Redux "out of the box" ... if you're happy with this I'll use these templates while creating a demo app and get them into better shape before asking for a merge.


ksnyde:
`.vscodeignore` is a default file that the vs-code editor adds to projects to help configure itself for the project. No harm in it being here but you're right it's not at all related to the PR.

ksnyde:
`.jsconfig` fits into the same category as another file that vs-code puts into each JS project ... personally I'd prefer to keep these two files in as it just makes anyone using vs-code life a bit easier but if it's too noisy I can remove.

blueprints/reducer/files/__root__/reducers/__name__.js
```diff
@@ -0,0 +1,38 @@
+// __name__ REDUCER
```
ksnyde: mistakenly I had thought ember-cli would populate with the name but since it doesn't I've removed

blueprints/ember-redux/files/__root__/reducers/index.js
```diff
@@ -0,0 +1,18 @@
...
+ * a discrete part of the state tree.
+ *
+ * Use the "ember generate reducer" command to create additional
+ * reducers.
+ */
+
+// import foo from './foo';
+// import bar from './bar';
+
+export default combineReducers({
```
ksnyde: I would think you would want this as the primary reducer in anything but a small app will just be handing off (aka, combining) to other reducers.

jsconfig.json
```diff
@@ -0,0 +1 @@
+{"compilerOptions":{"target":"es6","experimentalDecorators":true},"exclude":["node_modules","bower_components","tmp","vendor",".git","dist"]}
```
ksnyde: it's one of the vs-code default "decorator" files as mentioned above ... if we're not going to have these in the addon then we should add them to the `.gitignore` ok?

blueprints/ember-redux/files/__root__/reducers/index.js
ksnyde: > **Note: ** wrt to the `ember.d.ts` you're seeing this because anyone who uses vs-code will also automatically get typings for Ember and so fits into the same/similar category to the other vs-code config files ... for this one I think we should definitely include `typings/*` in the .gitignore. My bad for letting this through.

ksnyde:
removing the `/app/reducers` files as they are redundant with the more flexible default index format which will work with both apps and addons.

blueprints/ember-redux/files/__root__/reducers/index.js
ksnyde: > That said, I've noticed the trend/pattern and think others would find this valuable. Overall it's likely I'll work up a simplified version of this pull request and get a simple/correct blueprint into the project.

I'm fairly sure the style that I'm using is consistent with the latest recommendations Dan Abramov. What about my style is "incorrect"?

ksnyde: Here's a screen shot from a recent presentation of his:



And with my annotations:

blueprints/reducer/files/__root__/reducers/__name__.js
ksnyde: Just reread your comment and realised maybe I didn't answer it the first time. The file `__name__.js` is the **ember-cli** way of abstracting the _name_ of a type of thing (aka, a "reducer" in this case). Combine that with the `index.js` at the root of the reducer generator and you get an abstraction for directory path between an "addon" and an "app".

ksnyde: arguably we could remove the APP / ADDON abstraction because ember-cli now supports the `--dummy` flag but the `__name__.js` is still the _right way_ to name generated files.

jsconfig.json
ksnyde: this has now been done

blueprints/ember-redux/files/__root__/reducers/index.js
```diff
@@ -5,7 +5,7 @@ import { combineReducers } from 'npm:redux';
  * one or more other reducers which take on responsibility for
  * a discrete part of the state tree.
  *
- * Use the "ember generate reducer" command to create additional
+ * Use the "ember generate reducer [name]" command to create additional
```
ksnyde: Ideally when you use the `ember g reducer xyz` or `ember d reducer xyz` it would automatically populate the `index.js` reducer in the same way that `ember g route` does today.

toranb:
@ksnyde regarding the "incorrect" statement I made - it was in reference to the combineReducers code I saw in the reducers/index.js file (the original pull request).

`I would think you would want this as the main/primary reducer -- in anything but a very small app -- will just be handing off (aka, combining) to other reducers.`

I'm still uncertain if we are in sync here or if I'm just not understanding your intention. At the core I wanted to say that combineReducers will not be something you ever do yourself inside a reducer file because the service takes care of this for you.

About the pattern you mention - the only difference I use in ember apps is to avoid the default argument and instead use the `||` (or operator) in the default because JSHint isn't friendly. If ember makes the leap to ESLint at some point in the future it may be something I'd change but at this time I try to keep the "ember new + jshint + qunit" story working without error.

app/reducers/optional.js
```diff
@@ -1,7 +0,0 @@
-//this is a simple pass thru that offers an extension point
```
toranb: why was this deleted here? I assume a change this drastic would be included in another PR followed (likely) by a breaking api change as this is supported in ember-redux 1x right now

ksnyde: Hmmm. Maybe we are talking cross-purposes. If so then my apologies. I've seen no mention of `optional.js` in the docs. To me this was just an "indicator" that one could have more than one reducer file (something that is _very_ common in any real app).

If we were to have it though it would be better to include it as part of the blueprinting system so that you're supporting both Apps and Addons. Or am I missing something?

ksnyde: yeah there's more going on in the service than i'd imagined ... guess I should have looked rather than assume. I still need time to _digest_ functional programming so give me a sec to fully grok your setup.

blueprints/ember-redux/files/__root__/reducers/index.js
jkusa: @ksnyde, I agree. The functionality you are speaking of was suggested and implemented [here](https://github.com/toranb/ember-redux/pull/20/files#diff-d5d915533b340ef3b6bfbe8aacdaa6c2R14).

app/reducers/optional.js
ksnyde: WRT to **jshint** limitations ... makes sense; have been away from jshint for a bit and had thought maybe it supported ES6 more fully now.

ksnyde: the other difference in your pattern is the use of `if` ... `else if` which might get quite long; in these situations I typically prefer `switch`. To each his own I guess.

toranb: @ksnyde if you look at the documentation site under configure you can see my mention (along w/ plans to remove it likely in a 2x branch) :)

http://www.ember-redux.com/configure/

ksnyde: So just to be clear, here _was_ my understanding. Given a state tree like so:

image

Your `reducers/index.js` would be:

````js
import { combineReducers } from 'redux';
import auth from './auth';
import identity from './identity';
import presence from './presence';
import messages from './messages';
import outstanding from './outstanding';

export default combineReducers({
  auth,
  identity,
  presence,
  messages,
  outstanding
});
````

This in turn would make the `service/redux.js` (commented out lines are changes):

````js
import Ember from 'ember';
import redux from 'npm:redux';
// import optional from '../reducers/optional';
import initialState from '../reducers/initialState';
import middleware from '../middleware/index';

var { createStore, applyMiddleware, combineReducers, compose } = redux;
var createStoreWithMiddleware = compose(applyMiddleware(...middleware), enhancers)(createStore);

export default Ember.Service.extend({
    init() {
        // this.store = createStoreWithMiddleware(optional(combineReducers(reducers)));
        this.store = createStoreWithMiddleware(reducers, initialState);
        this._super(...arguments);
    },
    getState() {
        return this.store.getState();
    },
    dispatch(action) {
        return this.store.dispatch(action);
    },
    subscribe(func) {
        return this.store.subscribe(func);
    }
});
````

ksnyde: this implies there is always a `index.js` and `initialState.js` file in reducers. The `initialState.js` file would by default export an empty POJO object.

ksnyde: And when you `ember generate` or `ember destroy` reducers your `index.js` file is managed for you (like the `router.js` file does with routes).

ksnyde: With your implementation I see that `index.js` has simply been reduced to the export of a JS index. Not exposing the `combineReducers` _might_ be a benefit but if it's a managed file then I would say it's looking like a detriment to clarity. I think most of your audience will be aware of the `combineReducers` part of the API (it gets talked about often) and may well use it downstream in their own reducers so obscuring it here may just be hiding an important detail.

What do you think?

> p.s. I'm still not fully grokking how your "optional.js" is working currently so apologies there

ksnyde: Currently because you're missing a default `index.js` the first thing a user experiences is an error:

ksnyde: Now if I add my own index reducer as:

````js
import foo from './foo';

export default {
  foo
};
````

And then add `reducers/foo.js` as:

````js
const reducer = (state = {}, action) => {
  return {};
}
````

it fails with the error:

image

ksnyde: Maybe I'm still missing something but the solution I'm proposing offers:

- works immediately after install ... reducer just returns the initial state but it works without error
- easily understandable way to add `initialState` which can be very useful; defaults to empty object
- allows a simple `ember generate` / `ember destroy` metaphor for extending the reducers

I did find your documentation around `optional.js` and I think this should be deprecated for a variety of reasons:

- it is not in any part a standard part of Redux
- it's naming is confusing and the fact that you have a non-reducer function in the `/reducers` directory is a bad idea
- you're own admission that you had added it in the first place because you hadn't fully grokked middleware when you added it

Functionally the combination of _initial state_ and _middleware_ should meet any use case that historically used _optional_ but as you point out removing it outright can only happen on a major version but I'd suggest making this section a footnote in the `config` section of your documentation and logging a deprecation when it's used.

ksnyde: I'll quickly mock this up here in the PR so you can see what I mean. Hope this is all coming in the way it was intended. I would love to contribute to this as I think getting Redux into Ember and making it easy for folks is pretty important.

ksnyde:
In order to be as clear as possible about this PR, i've created a short video which demonstrates it's use:

https://vimeo.com/183007436

ksnyde:
wanted to make sure all tests were fixed for your review today; if you have any questions you want to talk through today you can find me on Ember Slack so long as it's before 8pm UK/London time.

toranb:
@ksnyde I think the biggest struggle I've had with this pull request is that it started out as a very small "blueprint" addition and quickly scope creep'd itself into dumping 1x features, adding blueprints for a concept not yet realized or discussed. In addition, a few of the changes you mentioned won't be something I merge in the 1x or likely 2x version of this addon.

Your feedback in the video was very helpful so I'll layout the things I want, and the things I don't along with a path forward.

1) the `optional` reducer needs to die

I realized very early on that between middleware, enhancers and reducers I added something unnecessary and likely confusing for new people who read through the source. I plan to remove this myself with another pull request (likely in the coming week) as part of the 2x release I've been working on.

2) the blueprint you added for reducers is much needed

I like the actual blueprint you created that will add/remove a reducer (and how it alters the reducer/index.js file to include this). Truly this saves time and is a net win. I'd like this to be included in another pull request that is very focused -truly just that one addition

3) the `combineReducers` change isn't something I want in reducers/index.js

I realize you feel this is better explicit but truly even in my react apps I prefer a single reducers/index.js file that just returns an object (allowing the boilerplate for redux to exist out of that file). In the example react app below I show how I've used combineReducers in a similar way to what happens in the ember-redux service here.

https://github.com/toranb/redux-beginner-introduction/blob/master/index.js#L68

4) the default reducer (reducers/index.js) file is excellent

I'd like you to do a simple pull request outside of this PR that includes this 1 change. I like how it prevents the basic error and the basic docs are always helpful for a beginner

5) initializer feature you added still needs some examples/ research

I realize conceptually this is helpful for offline and that has me interested. But truly I need to see more example apps with this to see how it works/ and some time to soak on it. Because this is a new api I want to leave it out of the other 2 pull requests I've asked for above with it's own issue showing docs, examples from react and other information about the problem (starting with a basic problem statement, solution like approach)

6) in general - I don't want any changes to the service itself

I mention the 2x release briefly above and I've been working on performance improvements along w/ the removal of optional.js (the reducer). But in truth, this 2x release shouldn't be a breaking change for anyone using the project today (happy path -avoiding optional). The only reason I'm doing a true 2x release is that optional exists and removing it can't occur in 1x, but in addition the performance work could give more advanced users some reason to add additional computed keys (still just barely a breaking change in my view overall). All that to say, I really don't want to alter the very core of this addon for some time. I want people to use it and feel confident that nothing is changing. In truth this addon as it exists today is the basis for all the ember work I do personally and I don't see any "missing" features at the moment (minus getting a working demo that shows off the benefits of redux-saga, or immutableJS)

7) Avoid gitignore, typescript, and all the messy DS_Store files

I'll expect a much more concise and clean pull request going forward -meaning that you've got to take a hard look at all files touched (using git diff) before pushing up. For example, in this PR alone if you look at github it's showing some 22 files changes for what started out as a single blueprint addition (or default reducer). 

I hope my tone here is nothing but patient and kind. I'm using this pull request as education and mentoring in a lot of ways. I see your value to the project and wider ember community and don't want to upset you or send out any negative vibe here. But keep in mind that I'm running this project from a very selfish slant because any code that comes in is something I agree to maintain.

ksnyde:
I fully respect your wishes to do as you like with your repository. From my end it felt from the start there was a resistance and the pace (I disagree with "scope") of my work seems to have exaggerated. I also feel the words "patient and kind" a little bit derogatory although maybe that's just how they came across.

A few take aways for you:

- All typescript files were removed within 15 minutes of the first review. As I said then, the two options were to include them in `.gitignore` or include them in the repo. Was happy to go either way. This is trivial stuff.
- The .DS_Store files were an oversight but a non-issue to remove (and add to the .gitignore)
- Not sure why you're suggesting avoiding gitignore as this is a critical file _to_ update if you're ensuring editor/OS cruft be left out of the repo.
- no changes were made to the service ... not sure why we're talking about that
- the visibility of `combineReducers` is just a design consideration which I was attempting to bring up a discussion around but you should rightly be the decision maker on this; it would have been nice to hear _why_ you're holding onto your view but ultimately it's fine either way ... btw, the fact that you're using it React that way too is not relevant to the discussion. 
- The only real change which came into being during the course of the PR was the "state-initializers" ... it's usage is very common in Redux and the implementation is completely symmetric to the reducers (which you seem to agree with). My feeling was there was zero downside and a lot of upside to including it. If you want more examples of use-cases I'd suggest an excellent video from Dan Abranov: [javascript-redux-supplying-the-initial-state](https://egghead.io/lessons/javascript-redux-supplying-the-initial-state) (unfortunately I think it may only be for paying members but that whole series is very good and in my content worth paying for).

I have loved your work on this addon to-date and I've loved watching some of the content you've produced over time Toran but I feel your interaction on this PR was less than welcoming and I will bow out at this point. Feel free to use any parts of my PR if they're helpful. No attribution needed.

toranb:
@ksnyde great points all around. I'm sorry for what seemed like resistance and clearly I've failed at open source based on this interaction. Please keep doing what you do - I'm certain other open source maintainers will benefit greatly from my failure to communicate in this context :(

PR #23, opened 2016-09-19
Only re-render when truly required

toranb:
This pull request aims to solve the performance problem that exists today with subscribe. Currently on any change we ask each component to re-render everything (not ideal) so in this PR we only ask for a re-render when state watched locally is truly different.

### DIFF
--- a/addon/-private/equal.js
+++ b/addon/-private/equal.js
@@ -1,0 +1,25 @@
+// https://github.com/reactjs/react-redux/blob/master/src/utils/shallowEqual.js
+
+export default function shallowEqual(objA, objB) {
+    if (objA === objB) {
+        return true;
+    }
+
+    const keysA = Object.keys(objA);
+    const keysB = Object.keys(objB);
+
+    if (keysA.length !== keysB.length) {
+        return false;
+    }
+
+    // Test for A's keys different from B.
+    const hasOwn = Object.prototype.hasOwnProperty;
+    for (let i = 0; i < keysA.length; i++) {
+        if (!hasOwn.call(objB, keysA[i]) ||
+            objA[keysA[i]] !== objB[keysA[i]]) {
+            return false;
+        }
+    }
+
+    return true;
+}
--- a/addon/components/connect.js
+++ b/addon/components/connect.js
@@ -1,4 +1,5 @@
 import Ember from 'ember';
+import shallowEqual from '../-private/equal';
 
 const { computed, defineProperty } = Ember;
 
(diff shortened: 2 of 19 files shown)


app/-private/equal.js
```diff
@@ -0,0 +1 @@
+export { default } from 'ember-redux/-private/equal';
```
dustinfarris: Why do we want private logic included in `app/`?

addon/-private/equal.js
```diff
@@ -0,0 +1,25 @@
...
+  // Test for A's keys different from B.
+  const hasOwn = Object.prototype.hasOwnProperty
+  for (let i = 0; i < keysA.length; i++) {
+    if (!hasOwn.call(objB, keysA[i]) ||
+        objA[keysA[i]] !== objB[keysA[i]]) {
+      return false
+    }
+  }
+
+  return true
```
dustinfarris: No semicolons in this file (I realize this was a copy+paste from react-redux).  Do we care about consistency in this regard?

app/-private/equal.js
toranb: not excited about it but I did need the ability to unit test it (and couldn't require this w/out it being in app with an export)

addon/-private/equal.js
toranb: fixed the formatting :)

dustinfarris: :+1:

app/-private/equal.js
dustinfarris: unit tests still pass for me when I remove this file

toranb: I was able to pull this :) thanks for the call out

addon/components/connect.js
```diff
@@ -38,14 +39,31 @@ var connect = function(mapStateToComputed, mapDispatchToActions) {
...
+                    this.unsubscribe = redux.subscribe(this.handleChange.bind(this));
                 }
                 this._super(...arguments);
             },
+            handleChange() {
+                var redux = this.get('redux');
+                var props = mapState(redux.getState());
+                var componentState = this.getComponentState(props);
+                var reduxState = finalMapStateToComputed(redux.getState());
+                if (!shallowEqual(componentState, reduxState)) {
```
foxnewsnetwork: <del>This is a great implementation and thanks for keeping up the good work on this, but is there anyway you can do something like:</del>
```javascript
handeChange() { 
  var redux = ...
  ...
  const isEqual = this.getWithDefault('stateEqual', shallowEqual);
  ...
}
```
<del>And thereby expose an api for customizing our equality checker. This will be particularly useful for a lot of us who are using [ember-immutable](https://www.npmjs.com/package/ember-immutable), [seamless-immutable](https://www.npmjs.com/package/ember-seamless-immutable), or maybe even some custom library that ships with its own equality checker.</del>

edit: never-mind, I'm stupid, lol. You're already putting state into flat pojos

PR #25, opened 2016-10-05
Add option for middleware setup

cafreeman:
Some Redux middleware require that additional setup occur after the middleware and store have been created. For example, to use [redux-saga](http://yelouafi.github.io/redux-saga/index.html), you have to start each listener once you've created the store and applied the Saga middleware, e.g.:

```js
// create the saga middleware
const sagaMiddleware = createSagaMiddleware()
// mount it on the Store
const store = createStore(
  reducer,
  applyMiddleware(sagaMiddleware)
)

// then run the saga
sagaMiddleware.run(mySaga)
```

In the current `ember-redux` setup, users do not have an opportunity to do this additional middleware setup after creating the store, since the store/middleware setup is handle in the `init` method of the `redux` service.

This PR adds a new option for middleware configuration that allows users to create a `setup` function that contains any setup code that needs to be run after the store is created.

In order to consume this new API, users can export a hash from `middleware/index.js` that contains the array of middleware (the same as what they do now), and a `setup` property that contains a thunk which runs the setup code.

For example:
```js
// middleware/index.js
sagaMiddleware = createSagaMiddleware.default();

const setup = () => {
  sagaMiddleware.run(someCoolSaga);
};

export default {
  middleware: [ sagaMiddleware ],
  setup
};
```

When the user exports a has (like in the above example), the `redux` service will extract the setup thunk and execute in the `init` method of the service, after creating the store.

It's important to note that this change is backwards compatible. The user may also still export an array of middleware, since the `redux` service checks for both cases and adapts accordingly.


toranb:
@cafreeman excellent! I originally worked around this using ember.run.next but ... as any hack goes :) it feel down fast. For those who might want to see a working redux-saga example checkout the basic one below (not doing much yet as we need a true api like you've thrown up here)

https://github.com/toranb/ember-redux-saga-example/commit/600fd300edf458d1959ba212ddafa109235c538f

About the api changes you threw up here - I can't say I know of anything more elegant. I appreciate the effort you put in to keep it backwards compatible as I'm doing everything in my power to keep 1x stable until April 2017 (assuming we don't bump up against something that forces 2x sooner).

I'll let others watching this give it a look over the next few days if that's cool with you. I plan to think about how one might test this configuration point from the acceptance level (assuming I can monkey patch something /modify the middleware for the dummy app on the fly to exercise both redux-thunk and redux-saga setups).

Thanks again for the work on this! redux-saga seems to be the clear winner once you outgrow redux-thunk so I look forward to supporting both thanks to your hard work!

app/services/redux.js
```diff
@@ -3,14 +3,32 @@ import redux from 'npm:redux';
 import reducers from '../reducers/index';
 import enhancers from '../enhancers/index';
 import optional from '../reducers/optional';
-import middleware from '../middleware/index';
+import middlewareConfig from '../middleware/index';
+
+const { assert, isArray } = Ember;
+
+// Util for handling the case where no setup thunk was created in middleware
+const noOp = () => {};
```
dustinfarris: You could use [`Ember.K`](http://emberjs.com/api/#method_K) instead

cafreeman: Good point! I went ahead and pushed that up.

cafreeman:
Thanks guys! I've addressed @dustinfarris feedback and also pushed up a change that passes the store into the `setup` function in case you need to access it. I ran into this with redux-observable because you need to do some additional setup to get the dev tools working.

If you want to see what this looks like in action, I've got a repo that's a hodge podge of ember-redux things located [here](https://github.com/cafreeman/github-dashboard-redux). Right now it's using Immutable, Redux-Saga, and (a very simple PoC of) Redux-Observable. I plan to keep experimenting with it as I continue to learn about the redux ecosystem

toranb:
@cafreeman thanks for the quick feedback and changes. I really like your example and the forward thinking you've had by looking into both saga and observable. I think the only part remaining is to understand better how we can test a few of these configurations at once using the dummy app. I've got this in my top 3 (short list) and I'll plan to see what all we can make happen!

toranb:
@cafreeman wow ... 8 days ago since you submitted this and I've done nothing to help merge it. Sorry for the delay my friend - long weekends usually hurt more than help these days but I'm looking at "how to test drive it" tonight so wish me luck!

toranb:
@cafreeman got a test up for it - look this over when you get time this week. If you approve, we can have you rebase after my merge 

https://github.com/toranb/ember-redux/pull/26

krivaten:
Looks solid on my end. Great work @cafreeman!

cafreeman:
@toranb no worries, friend :) I don't mind the wait. That test looks good! Looking forward to rebasing and seeing the tests go green!

toranb:
@cafreeman rebase at will my friend!

cafreeman:
Looking good!

screen shot 2016-10-14 at 8 34 50 am

cafreeman:
All green!  🙌

toranb:
@cafreeman I just published v1.7.0 w/ your changes

Thanks again for the time you invested to help land this! I'm planning to update the gh-pages/docs website over the weekend to show how you can use this (by example) with redux-saga

PR #26, opened 2016-10-14
[TESTS]: wrote a full redux-saga example w/ middleware

toranb:


toranb:
@cafreeman I wrote a simple patch that helps prove redux-saga is supported (along w/ your custom middleware setup method). I realize it doesn't include any proof that the store is passed like redux-observable needs but I figure we could patch this in w/ a dual middleware (like you showed in your example).

This look good to you? It's not a "running test" just yet as I planned to merge this (after you look it over) and then you can rebase/ rename it to middleware-test.js and watch it go green :)

cafreeman:
@toranb this looks great! Let's merge it!

PR #30, opened 2016-10-27
Cache the mapped state in components.

brettburley:
I noticed that the `mapStateToComputed` function is called more often than needed (in fact, it was called once for every single property it provides), and depending on the type of operations done to map the state, this can sometimes be expensive.

I also considered another approach where I simply do:

```
redux.subscribe(() => {
  const state = finalMapStateToComputed(redux.getState());
  this.setProperties(state);
});
```

which I like because it's very simple, and takes advantage of Ember's no-op on `set` if the value is the same (and thus no dirty checking required). However, doing this loses the ability to mark these properties as "readonly". But I think it's still worthy to consider.

---

This prevents the mapStateToComputed function from being called for
every single property that it provides to the component. Additionally,
this change also only notifies for properties that have actually
changed their value to reduce the amount of rendering required when
there are changes.


brettburley:
I forgot to mention that I also changed the `rerender-test` to count the actual number of re-renders via Ember's `didRender` hook. The number of total re-renders for each component is identical pre- and post- the change to cache the local mapped state & notify only for properties that have changed.

toranb:
@brettburley great catch - I noticed this just the other day and thought to myself "what now ??"

I need to pull this down and experiment with it - the biggest fear I have looking over the commit is the fact that you are now holding/and updating state (within the extended component) -something I avoided on purpose. That said, it came at a cost and I'll need to confirm this is the right path forward. Overall- great work here @brettburley !!!

brettburley:
Yeah, a few things come to mind:
1. It looks like react-redux stores some local state in each component as well. It has a pointer to both `storeState` as well as to `stateProps` for the computed state: https://github.com/reactjs/react-redux/blob/master/src/components/connect.js
2. On the topic of following react's implementation vs. optimizing for Ember, it seems the later would seem to be preferable. That way, `ember-redux` can optimize it's implementation to work best with Ember and its rendering engine.

Definitely do take this for a spin as well -- any kind of optimization like this always has a risk of error.

toranb:
@brettburley some early feedback - I still think the shallowEqual is useful/ valuable. When I switched the test back to "how many times do we slip past the equality check" I see a measurable drop (from 6 to 1 for the bottom 2 items in that acceptance test) using that shallowEqual you removed. note: this was to show how often we get beyond the return line (not how many re-renders are actually invoked inside ember)

I realize you moved the metric to didRenders but if we can prevent the glimmer engine from having to do any type of work isn't that a win? In this case I simply swapped your (oldstate === newstate) for something like this

```js
if (!shallowEqual(oldState, newState)) {
}
```

brettburley:
@toranb where were you counting "slip past equality check" on this branch? Was it at the `oldState === newState` line?

I believe that line 59 https://github.com/toranb/ember-redux/pull/30/files#diff-cc792737be606b77161ace89ad8a3cb4R59: `if (cachedValue !== undefined && cachedValue !== newState[key]) {`, should be doing something very similar to what the shallow equal check was doing previously?

toranb:
@brettburley I threw up a quick n dirty branch to show this change off

https://github.com/toranb/ember-redux/commit/6df5d7016e30a21234ee40c791879491c047bc12

I wanted to show that the shallowEqual function has value. Using that instead of the 2 conditionals you mention results in this result

```js
        assert.equal(oneUpdated, 4);
        assert.equal(twoUpdated, 4);
        assert.equal(fourUpdated, 1);
        assert.equal(fiveUpdated, 1);
```

And when I ran this same set of assertions with the `oldState === newState` check I got many more calls (the original behavior before my PR that added shallowEqual to begin with).

```js
        assert.equal(oneUpdated, 6);
        assert.equal(twoUpdated, 6);
        assert.equal(fourUpdated, 6);
        assert.equal(fiveUpdated, 6);
```

I'm not suggesting your work isn't great here - I'm mostly hoping you can fact check me/ and we could combine efforts to get the best performance result. Ideally I'd like to use shallowEqual to reduce the number of calls -and your work (using some local state in the connected component) to reduce the amount of calls via `mapStateToComputed`

I still plan to look over this PR further but if you have time this week I'd like to see if you can confirm my findings (that shallowEqual is useful here) but it's possible my little experiment didn't pull over your connect changes as desired :)

toranb:
@brettburley disregard those bottom numbers for a day or so - I realized last night that I wasn't tracking this accurately :)

I'll hack on this again tonight and see what I can learn -sorry for the distraction!

toranb:
@brettburley Sorry for the delay, I had trouble getting my monkey patch to work 

```js
assert.equal(oneUpdated, 8);
assert.equal(twoUpdated, 4);
assert.equal(fourUpdated, 1);
assert.equal(fiveUpdated, 1);
```

From what I can see at runtime, the cache we are using never results in equality meaning we always notifyPropertyChange. Here is a snap of the 2 arrays that should be equal but are not.

screen shot 2016-11-02 at 8 29 13 am

This brought to my attention the fact that we are now using `cacheFor` in the equality check itself. Would it be worth considering some of what you have here + some of what I have (using the shallowEqual) as it seems the shallowEqual approach does indeed improve efficiency w/out sacrificing integrity?

toranb:
@brettburley I pushed a revised look to that branch showing how we could incorporate both your improvements while remaining to keep the connected components as efficient as possible

https://github.com/toranb/ember-redux/commits/branches/Alternative

brettburley:
Good find @toranb. I definitely did not expect the cachedValue to not be the same as the new state property.

With regards to `shallowEqual`, I think it has a lot more utility with React than it would in Ember. In React, if any of the state changes, then the entire virtual dom has to be regenerated so that it can be diff'd. So in React, they only need to answer "has anything changed?" On the other hand, in Ember, the rendering engine needs only to render parts of the template that depend on a particular property, so the question becomes "what has changed?" I was trying to take advantage of this by only calling `notifyPropertyChange` for properties that have actually changed (rather than aggressively calling it on all properties). I think it's advantageous for us to go a step further than `shallowEqual` and actually check what properties have changed.

That said, it looks like the comparison for each property is not working as I had expected, so I definitely want to look into this.

toranb:
@brettburley About the shallowEqual - I'm totally in agreement that ember's view engine (glimmer) will do most of the heavy lifting as you mention, but I don't feel we should notifyPropertyChange if it's not truly required. The point you mention that has me thinking here is - what if we map 8 props, but only 1 has truly "changed" - in that scenario we should only notifyPropertyChange that 1 prop (not the other 7). Is that in part what you are making comment about?

As you mention, something isn't right here because the local state (from redux) is a POJO, but the cache value is a ember wrapped POJO and won't ever be equal. Are you planning to look at this further in the coming week(s) ?

brettburley:
> Is that in part what you are making comment about?

Yes, precisely! For the highest performance, we should only notify for properties that have actually changed. Notifying all properties in `newState` when `!shallowEqual(oldState, newState)` will over-notify. So we need to check the equality of each property so that we can only notify when a property is changed. But if we're doing that, then there's no use in using the `shallowEqual` function directly because that's exactly what it does internally (checks equality of each property): https://github.com/toranb/ember-redux/blob/28ad7d1b76c418010d94157293bbfe77e0a199d8/addon/-private/equal.js#L19

And is what I have replaced with the check here: https://github.com/toranb/ember-redux/pull/30/files#diff-cc792737be606b77161ace89ad8a3cb4R59

I counted the number of times `notifyPropertyChange` is invoked in total in the rerender test. On this branch, it's invoked 14 times. If I only use the `shallowEqual`, then it's invoked 18 times in total. On master it's invoked 29 times for the same test! Here's an example case where the equality check succeeded, preventing an unnecessary `notifyPropertyChange`:
screen shot 2016-11-02 at 9 04 24 pm

Regarding the behavior you saw with `cachedValue` having its objects decorated by Ember, it looks like Ember is in fact mutating the POJOs that are provided to it in the computed props. I checked out `master`, and it looks like it happens even there:
screen shot 2016-11-02 at 9 18 45 pm

I think this is a separate issue that we should take a look into eventually as well and see if there's anything to optimize or do better there.

brettburley:
FYI, it looks like the object decoration happens at render time by Ember when it sets up observers (for example, for the key `name` on an item in the `items` array): https://github.com/emberjs/ember.js/blob/v2.8.2/packages/ember-htmlbars/lib/streams/key-stream.js#L67

Can't imagine there's much we can do about this right now.

toranb:
@brettburley you are correct -let's skip the addObserver work as it's not likely in scope right now. Can we keep the test as it was - to help us confirm the number of `notifyPropertyChange` events we see instead of didRender? The intent of the test was to lock down and confirm the number of times we call this.

Could we also update the test cases such that one of the views we see updated each time has a new property that isn't updated? This would help seal the deal around your value-add (that we never over-notify)

Also -as you have it, I'd like the shallowEqual completely removed :)

dustinfarris:
What about a hybrid approach that stores an private property for each prop, and then a oneWay for the actual user-facing prop??

e.g. (pseudo-ish)

```js
getPrivateKey = (key) => {
  // Underscore prefix
  return `_${key}`;
}

getPrivateState = (propState) => {
  return Object.keys(propState).reduce((privateState, key) => {
    return Object.assign(privateState, { [getPrivateKey(key)]: propState[key] });
  }, {});
}

setPrivateState = (state) => {
  const propState = finalMapStateToComputed(state);
  component.setProperties(getPrivateState(propState));
}

init() {
  setPrivateState(redux.getState());

  Object.keys(propState).forEach(key => {
    defineProperty(component, key, Ember.computed.readOnly(getPrivateKey(key)));
  });

  redux.subscribe(() => {
    Ember.run(() => {
      setPrivateState(redux.getState());
    });
  });
}
```

I haven't tested this yet.  Just thinking out loud.

toranb:
@dustinfarris I've spent a great deal of effort trying to avoid additional state in the component itself because when you have state duplicated you often pay a tax to keep that state in sync at some point (or it can cause confusion about which "state" is the correct source of truth).

I think we've got 2 issues here and 1 that I'm planning to solve tonight myself is the "don't over notify" problem @brettburley mentioned above. That is the single biggest performance issue I see here right now and one that people are taking note of outside of this PR.

The 2nd issue is that of "can we call mapState just 1x instead of 2x" and I want to spend more time thinking about this one. It has room for improvement as @brettburley mentioned above, but I the tradeoff of storing state locally to prevent duplicate calls. As I'm trying hard to avoid this, I'd like to spend more time talking about that problem directly.

dustinfarris:
@toranb good points, and i'm not sure at all what the right way forward is.

But here are some more of my thoughts for the sake of seeing all the edges of the argument and a possible solution somewhere in this PR.

Bouncing off the issues you raised:

**Issue (0): Avoid unnecessary component state** — Whichever way we do this, there is going to be a set of component properties as specified by the user.  That is inescapable.  Adding them as private properties, and then adding public "readOnly copies" does not add additional state because `Ember.computed.readOnly` is actually just `Ember.computed.alias` with some extra rules.  So we are not duplicating state, only creating internal logic that allows us to connect the redux state to the component props in a way that is efficient and flexible for us.

**Issue (1): Don't over-notify** — Using `Ember.Object.set` (and, by extension, `Ember.Object.setProperties`) will noop (as in, no `propertyDidChange` fired, which means observers are not notified) if the old and new values are equal.  @brettburley has pointed this out as well.  We could leverage Ember's exsiting optimization here.

**Issue (2): Call mapState just once** — the hybrid method I've postulated (as an evolution of the much simpler version in the PR description that @brettburley put forward) only calls mapState (or rather my "private" version of that) once per state-change

toranb:
@brettburley @dustinfarris I've got a PR up to handle the over notify problem mentioned -please give this a look when you have time.

@brettburley assuming we merge that, could we shift the focus of this PR (or open a new one?) that is solely focused on the mapState/how best to use (local) component state problem mentioned above?

toranb:
@brettburley can we close this out now that @dustinfarris has taken a run at the setProperties approach and I've handled the over notify problem?

brettburley:
Yep! Closing now.

PR #31, opened 2016-10-27
Call super in component willDestroy.

brettburley:
So that the base component can use the `willDestroy` hook too.

### DIFF
--- a/addon/components/connect.js
+++ b/addon/components/connect.js
@@ -65,6 +65,7 @@ var connect = function(mapStateToComputed, mapDispatchToActions) {
                 });
             },
             willDestroy() {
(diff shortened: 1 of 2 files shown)


toranb:
@brettburley totally had this on my list :) but first I wanted to create a test scenario that shows this is truly needed/required. You interested or should I take a swing at it in the near future/to help land this pull request?

brettburley:
@toranb thanks for calling me out on adding a test. There definitely should be one.

Just added one, please take a look when you get a chance.

toranb:
@brettburley quick work my good man! thanks for the great test case also :)

PR #32, opened 2016-10-28
Use setProperties to propagate redux mapped state to components.

brettburley:
I wanted to share this to get your thoughts. It would be an alternative to https://github.com/toranb/ember-redux/pull/30. I love how simple it is, taking advantage of Ember's noop on set when the values are the same. While state is duplicated locally, I think there is little risk for it getting out of sync because it's so simple.

However, I think that the biggest downside is that this loses the ability to have the `readOnly` annotation on the properties, so it would be possible for a consumer to `set` them. Given that this is also the case when a component is passed in properties in Ember (`{{my-component prop=foo}}`, the component can do a set on `prop` which propagates to `foo`), I think it's not the worst thing in the world, and certainly something to revisit once Glimmer components are out and immutable props exist.

---

This allows Ember to optimize on set and only set each property when
it has changed. It also only requires a single execution of
mapStateToComputed each time the store's state changes.


toranb:
@brettburley I do actually respect the simplicity here but in truth I'm fairly dogmatic about the readOnly property. It's helps to eliminate an entire class of defect that has trolled me/ and I'm certain countless others :)

brettburley:
Yeah, I had two thoughts about this. First, there's unfortunately no protection against mutating objects that exist within properties. For example, if `mapStateToComputed` returns `{ foo: [1, 2, 3] }`, the component can still do `this.set('foo.firstObject', 5)` and mutate the array. However, to this point I do think that it's more likely that you might accidentally try to do `this.set('foo', [])`, where the readOnly protection would help. :+1:

Secondly, folks who are more familiar with Ember might really value the extra safeguards. I would even suggest, perhaps, using a custom setter that throws an exception with a more domain-specific message like:
> Exception: dispatch a redux action to mutate state rather than calling `set` directly.

toranb:
@brettburley I'm still trying to understand if having this PR open adds value or if we should close it/focus in on the other (that respects readOnly) -thoughts?

toranb:
@brettburley closing this out for now - always open to the idea of another PR centered around additional guarantees/protection but it seems we #30 in flight and that's likely the winner here

PR #35, opened 2016-10-31
[BUG]: wrap updateProps in Ember.run

dustinfarris:
If the redux state is changed by an outsider, acceptance tests can potentially fail as rerender occurs after the test completes.

Wrapping updateProps with Ember.run fixes this.

### DIFF
--- a/addon/components/connect.js
+++ b/addon/components/connect.js
@@ -1,7 +1,7 @@
 import Ember from 'ember';
 import shallowEqual from '../-private/equal';
 
-const { computed, defineProperty } = Ember;
+const { computed, defineProperty, run } = Ember;
 
 var connect = function(mapStateToComputed, mapDispatchToActions) {
     var shouldSubscribe = Boolean(mapStateToComputed);
@@ -44,13 +44,15 @@ var connect = function(mapStateToComputed, mapDispatchToActions) {
                 this._super(...arguments);
             },
             handleChange() {
-                var redux = this.get('redux');
-                var props = mapState(redux.getState());
-                var componentState = this.getComponentState(props);
-                var reduxState = finalMapStateToComputed(redux.getState());
-                if (!shallowEqual(componentState, reduxState)) {
-                    this.updateProps(props);
-                }
+                run(() => {
+                    var redux = this.get('redux');
+                    var props = mapState(redux.getState());
+                    var componentState = this.getComponentState(props);
+                    var reduxState = finalMapStateToComputed(redux.getState());
+                    if (!shallowEqual(componentState, reduxState)) {
+                        this.updateProps(props);
+                    }
+                });
             },
             getComponentState(props) {
                 var componentState = {};
--- a/tests/acceptance/rerender-test.js
+++ b/tests/acceptance/rerender-test.js
@@ -13,6 +13,23 @@ module('Acceptance | rerender test', {
     }
 });
 
+test('should rerender when state is changed by other redux consumers', function(assert) {
+  const redux = application.__container__.lookup('service:redux');
+  ajax('/api/lists', 'GET', 200, []);
+
+  visit('/lists');
+
+  andThen(() => {
+    assert.equal(find('.list-item-one .item-name').length, 0, "No items initially");
+
+    redux.dispatch({type: 'TRANSFORM_LIST', response: [{id: 1, name: 'bob', reviews: [{rating: 5}]}]});
+
+    andThen(() => {
+      assert.equal(find('.list-item-one .item-name').length, 1, "One item after redux dispatch");
+    });
+  });
+});
+
 test('should only rerender when connected component is listening for each state used to compute', function(assert) {
     ajax('/api/lists', 'GET', 200, [{id: 1, name: 'one', reviews: [{rating: 5}, {rating: 5}]}, {id: 2, name: 'two', reviews: [{rating: 3}, {rating: 1}]}]);
     visit('/lists');


toranb:
@dustinfarris we've got a ton of async tests in this project (and I've used a good amount in ember apps (not just the addon dummy app here) and none of them require this as the render seems to be run loop aware. Could you boil down your example and trap the stack trace as this error is thrown to better understand/reproduce a test that would require this change?

dustinfarris:
I think it stems from doing async work in mapStateToComputed.  I'll work on a test to reproduce.

dustinfarris:
@toranb See above 3 commits.

I reverted the fix, wrote a test showing the problem, and added the fix back.

I've updated the description of this PR as well.  Let me know if this works for you and I'll squash.

toranb:
@dustinfarris ah this is why you ran into that issue :) 

Instead of changing production code to make this test friendly, use the ember async helper api like I show in this branch

https://github.com/toranb/ember-redux/commit/86b2237338e2fae2717ac6f2ce76caa16011c8f1

Also helps you avoid nested andThens :)

dustinfarris:
@toranb what about when you want to use upstream non-Ember libraries that are dispatching to the store?  (my case)

I think this is a perfect example of when you would want to wrap in Ember.run.

From the API docs:

> if you are implementing raw event handlers when interfacing with other libraries or plugins, you should probably wrap all of your code inside this call.

We are interfacing with redux and relying on events coming from state changes.  Absolutely makes sense to me.

toranb:
@dustinfarris good point - I'm curious how we could better simulate that here to express the real world scenario. I'm most curious also because I'm curious if we should/could try to join an existing run loop if one is already in motion (using ember.run.join I believe it is).

Do you have a branch or gist of the library that calls this? I'm curious what that looks like (hope you don't mind my asking)

dustinfarris:
@toranb not at all

I am using Cashay, which provides a nice interface to run GraphQL queries as your primary store interface.

Basically, when I want data, I do something like:

```js
let { users } = cashay.query(`{ users { id, name, age } }`).data;
```

or perhaps a more relevant graphql example might be:

```js
let { items } = cashay.query(`{ items { id, name, price, reviews { id, rating, comment } } }`).data;
```

Either way, cashay will check redux for the data.  If it's not there, it fires off a network request to your backend requesting only the data that's missing.  It updates redux if necessary, and returns the data to you.

This is done async, so the initial result from the `.query` call is an object that has a `status: "Loading"` attribute, but I digress...

To wire up Cashay, you create the redux store that includes a special reducer that is provided by the library.  Thanks to ember-redux, it is super simple for me to get a store put together https://github.com/dustinfarris/ember-cashay/blob/master/blueprints/ember-cashay/files/app/reducers/index.js

After creating a store that includes cashay's reducer, you create a cashay singleton, providing the store, and some other details such as the network transport.  You can see how ember-cashay does that here: https://github.com/dustinfarris/ember-cashay/blob/master/app/instance-initializers/ember-cashay.js#L12

From that point on, i'm just using `cashay.query` without coming into personal contact with redux.

Here is an example using an ember-redux-style component: https://github.com/dustinfarris/ember-cashay/blob/master/tests/dummy/app/components/users-list.js

The goal of ember-cashay is not to hide ember-redux, but to allow the user to use cashay and ember-redux side-by-side.  My vision is to use cashay for most persisted state, and layer on additional redux for any remaining application state, all wrapped up in `stateToComputed` and `dispatchToActions` handlers as exemplified by ember-redux.

dustinfarris:
@toranb i'm not sure if i actually answered your question there.  here is a place where cashay calls dispatch: https://github.com/mattkrick/cashay/blob/37daaea5f2056cdde851e64f21f04a6fb936e2b5/src/Cashay.js#L440

toranb:
@dustinfarris this is a great intro to the "what/why" - I'll need a little time to digest it but so far -so good :)

brettburley:
@dustinfarris, I'm really excited to see your work on `ember-cashay`!

It seems like there's three candidates for where integration into the run loop can happen:

1. In the application/addon/etc. code. This is similar to using plain jquery ajax in Ember today, e.g.: `$.getJSON(url).then(data => Ember.run(() => this.set('foo', data)));`.
2. At `dispatch` time. Conceptually you could imagine the `redux` service's `dispatch` method is rewritten to be `Ember.run.once(() => this.store.dispatch(action))`. This is basically what happens already when `dispatch` is called from an action.
3. In the `connect` component when handling store updates. This is what @dustinfarris proposed here.

Option (1) seems to be consistent with Ember, and makes dealing with async in `ember-redux` like dealing with async in Ember (dispatch would parallel Ember.set). However, it requires 3rd party libs that rely on redux to allow you to hook into that behavior, which I bet a lot don't have since react doesn't have these same async requirements.

Option (2) is my least favorite because it makes `dispatch` async since it will happen later in the run loop, which makes it harder to use its return value for model hooks, etc. It also adds behavior to `dispatch` that doesn't apply to the redux store's `dispatch` method. This again would be an issue for 3rd party libs that work directly on a redux store, without this addon patching the store's `dispatch` method.

Option (3) seems like it might be a reasonable way to approach the problem. It basically is saying that `redux` is external and may send updates at any (unknown) time, and therefore must be synchronized. I think `Ember.run.once` or `Ember.run.scheduleOnce('action', ...)` (the first is shorthand for the later) is probably the right method since it will use an existing run loop if it exists and will schedule it as part of the "actions" part of the run loop.

I'm actually not entirely sure why this is occurring in the `ember-cashay` case. Ember's RSVP.Promise is set up to synchronize with the run loop when `resolve` is called, and it looks like the HTTPTransport you defined is using Ember.RSVP.Promise: https://github.com/dustinfarris/ember-cashay/blob/master/addon/http-transport.js#L46, and I would imagine the `dispatch`ing from cashay would by sync with the transport resolving? I'm not sure if it's a weird interaction with async/await that cashay uses that causes it to be executed outside of Ember.RSVP.Promise's patched resolve.

dustinfarris:
@brettburley thank you for that great breakdown— i'll point out that the failing test in this PR simply executes `store.dispatch`.  No network activity or anything else going on.  I think your explanation of Option (3) is spot on.

However, I don't think `Ember.run.once` is the right way in this case.  `run.once` depends on a run-loop being automatically generated which, according to the docs, is inefficient in practice, and disabled completely in the test environment.

From [the guide](https://guides.emberjs.com/v2.9.0/applications/run-loop/#toc_how-is-run-loop-behaviour-different-when-testing):

> Autoruns are Embers way of not punishing you in production if you forget to open a run loop before you schedule callbacks on it. While this is useful in production, these are still situations that should be revealed in testing to help you find and fix them.

Indeed, the test fails if I change the code to `run.once(`.

I continue to believe that the implementation in this PR is the right way forward.

brettburley:
@dustinfarris good catch, I think you're probably right about using `Ember.run()`.

As for "simply executes `store.dispatch`", here's an example test case that also fails with the same error because there the `foo` object has a property set outside of the run loop (i.e. in the testing context):
```
test('prop set outside of run loop', function(assert) {
    this.register('component:my-component', Ember.Component.extend({
        layout: hbs`{{foo.prop}}`
    }));

    this.render(hbs`{{my-component foo=foo}}`);

    const foo = Ember.Object.create({});
    this.set('foo', foo);

    foo.set('prop', 'test');

    assert.expect(0);
});
```

To fix this, the line `foo.set('prop', 'test');` should be wrapped in a run: `Ember.run(() => foo.set('prop', 'test'));`. I think `dispatch`'s currently behavior is consistent with this. _If_ this failing test case was the only issue at hand, I would suggest wrapping the `dispatch` in an `Ember.run` in tests, or a test helper created, similarly to how Ember sets up `this.set` in component tests:
https://github.com/emberjs/ember-test-helpers/blob/master/lib/ember-test-helpers/test-module-for-component.js#L259

However, this obviously doesn't solve the 3rd party lib integration issue.

toranb:
@dustinfarris I've looked this over and think we have 2 options

1) use the ember.run as you've setup in this pull request
2) ask the people using dispatch to wrap that call in ember.run

For example, why not wrap your use of dispatch in ember.run? Would that not solve the core issue you ran into? What are the pros/cons of you calling it vs adding that call to each dispatch in ember-redux (the core library) itself?

dustinfarris:
@toranb: as @brettburley and i have stated further up in this thread, the user does not always do the dispatch directly.  other non-Ember libraries may call dispatch in completely unforeseen ways.

i cannot think of any pro for requiring the user to figure out how to get any possible thing that might do a dispatch wrapped in a run-loop.

the con of requiring the user to do this, of course, is that now the user must figure out how to get non-Ember dependency libraries that perform dispatches to cater to the Ember run-loop

the pro of wrapping the async side-effects (the component updates) of redux _as recommended by the Ember docs_ is that the user does not have to jump through hoops to ensure that redux dispatches are properly queued, executed, and flushed.

the con of wrapping the component updates is... i can't think of any

----

ember-redux provides a way to make "connected" components.  ember-redux should support the changes to the component caused by redux.

dustinfarris:
@toranb remember that virtually everything gets loaded into a run-loop in Ember _already_.  we are not doing anything more expensive than the norm here.  the only reason this isn't in a run-loop already is because it is coming from an event outside of Ember's knowledge (from a redux subscription handler)

again, from the docs:

> you should wrap any non-Ember async callbacks in Ember.run

if anything, we should probably wrap all of `handleChange` in a Ember.run instead of just `updateProps`

toranb:
@dustinfarris great points - looks like a merge tomorrow unless someone else has feedback! Thanks for taking the time to write - appologizes for my defensive tone

dustinfarris:
@toranb thanks.  I moved the call to wrap the entire callback—to keep in line with what the docs recommend.  I've also squashed, and updated the commit message to summarize the reasoning behind this change.

This should be good to go.

PR #39, opened 2016-11-05
Refactor `connect`

dustinfarris:
This is a different approach to the concerns raised in #30 + some additional refactoring.

Since this is a refactor, it might be easier to reason about by [looking at the new connect.js](https://github.com/dustinfarris/ember-redux/blob/617155000cb42267d081b59e41d2b82f0f51a2bc/addon/components/connect.js) rather than trying to make sense of the diff.

### Using `Ember.Object.setProperties`

This greatly simplifies the code and transfers optimization responsibility to Ember, where it should stay if at all possible.

The properties work like this:

On init (and after every state change via redux.subscribe), the user props are computed using the user-supplied `mapStateToComputed` function.  The resulting object might look like e.g.:

```js
// props
{
  users: [ { id: 1, name: 'Bob' } ]
}
```

The resulting props object is transformed into an equivalent object with "private" keys, e.g.:

```js
// props with private keys
{
  _users: [ { id: 1, name: 'Bob' } ]
}
```

The transformed object is set on the component using setProperties,

```js
this.setProperties(privateProps);
```

"read-only" aliases are set up to match what the user expects

```js
// Simplified
Ember.defineProperty(this, 'user', Ember.computed.readOnly('_user'));
```

This allows us to update the "private" component properties using redux.subscribe, and the user continues to consume the public read-only props (which are just aliases).  It is important to emphasize that _no additional state is being created_—just a pathway that allows us to make updates to the component and still keep everything locked down.

_Note: the `updateProps` counter tests are removed because updateProps now fires on every redux state change, as it should._

### Move logic to component methods

Rather than using behind-the-scenes functions, this PR moves most of the logic to a public API on the WrappedComponent object.  This feels a bit more Emberish to me, and more importantly, allows the user to adjust behavior if desired.

Specifically, these public methods are added (or modified):

- `WrappedComponent.getPrivateKey`
- `WrappedComponent.transformPropsToPrivate`
- `WrappedComponent.updateProps`

Most likely some unit tests are appropriate here.  I can add those, but would like feedback about this change first.  I think it is the right way to go, although we may want to make the methods private at first before committing to API support—and then make them public later on if it makes sense.

### Minor stuff

After discussing with @toranb I took some liberty to swap in ES6 syntax here and there.  Mostly this doesn't change anything, although I did set up default arguments to connect, and moved the logic for `shouldSubscribe` to a check in `init` for a length of props keys > 0.

### Dispatch not touched

Other than the default argument to mapDispatchToActions, I have left the dispatch logic alone.  Depending on how the refactor for properties is received, we may want to refactor dispatch to match the style and spirit _et al_.

## TODO

- [ ] Flesh out comments (we should probably use JSDoc format)
- [ ] Unit tests for WrappedComponent methods
- [ ] Decision to move mapToDispatch to component method

----

cc @brettburley

### DIFF
--- a/addon/components/connect.js
+++ b/addon/components/connect.js
@@ -1,70 +1,107 @@
 import Ember from 'ember';
 
-const { computed, defineProperty, run } = Ember;
+const {
+    computed: { readOnly },
+    defineProperty,
+    inject: { service },
+    run
+} = Ember;
 
-var connect = function(mapStateToComputed, mapDispatchToActions) {
-    var shouldSubscribe = Boolean(mapStateToComputed);
-    var finalMapStateToComputed = mapStateToComputed || function() {return {};};
-    var finalMapDispatchToActions = mapDispatchToActions || function() {return {};};
+export default function connect(mapStateToComputed = () => ({}), mapDispatchToActions = () => ({})) {
+
     return function wrapWithConnect(WrappedComponent) {
-        var mapState = function(state) {
-            var props = [];
-            Object.keys(finalMapStateToComputed(state)).forEach(function(key) {
-                props.push(key);
-            });
-            return props;
-        };
-        var mapDispatch= function(dispatch) {
-            var actions = [];
-            Object.keys(finalMapDispatchToActions(dispatch)).forEach(function(key) {
+
+        const mapDispatch = function(dispatch) {
+            let actions = [];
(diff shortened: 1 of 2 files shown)


dustinfarris:
Breaking this into separate PRs

Issue #37, opened 2016-11-04
Improving the notifying process


justinpark:
Hello @toranb 
Thank you for your solution.
Our team love your solution and try to apply into the internal solution.
I found one thing to fix to improve the observing performance.

I'm not a good writer. Please understand the lack of description.

The problem is the `connect` function notifies all mapped properties even if only one of them has been updated.

For example, I have one container component that watches the router state and a collection record.
```javascript
function select(store) {
  const {
    collection,
    router,
  } = this.store;
  return {
    collection,
    router,
  }
}
```
And then the component observes the router state and updates the collection set once the collection filter parameter is updated.
```
...
onRouteChange: Ember.observer('router', function() {
   const redux = this.get('redux');
   redux.dispatch({type: 'UPDATE_COLLECTION', payload: {filter: router.params}});
})
```
The router observer has been triggered after the collection state is updated because the `handleChange` compares the all states mapped and triggers the `notifyPropertyChange` all the states keys. This makes the infinite loop and breaks the system.
```javascript
                var props = mapState(redux.getState());
...
                if (!shallowEqual(componentState, reduxState)) {
                    //the following process notifies all props
                    this.updateProps(props);
                }
```

# Suggestion
We solved the problem by notifying the only updated state keys.
On `connect` side, we would change the notifying process only updated values.
```javascript
            handleChange() {
                var redux = this.get('redux');
                var props = mapState(redux.getState());
                var componentState = this.getComponentState(props);
                var reduxState = finalMapStateToComputed(redux.getState());
                props.forEach((name) => {
                  if (!shallowEqual(componentState[name], reduxState[name])) {
                    this.notifyPropertyChange(name);
                  }
                });
            },
```

How do you think of my idea?
If the suggestion works, please patch the change on the next version.

Thanks.

Regards,
@justinpark

toranb:
Good news :) we already have a PR in flight to solve the problem you mention

https://github.com/toranb/ember-redux/pull/30

toranb:
To push this ahead much quicker, I've opened a PR solely focused on the over notify problem you bring up here. 

https://github.com/toranb/ember-redux/pull/38

My hope is we can get a release out this weekend -sorry for the delay!

toranb:
@justinpark I just published v1.8.0 w/ this change. Please pull this down and confirm when you have time :)

PR #41, opened 2016-11-09
Use Ember.Object.setProperties to update props

dustinfarris:
This is a different approach to the concerns raised in #30

### Using `Ember.Object.setProperties`

This greatly simplifies the code and transfers optimization responsibility to Ember, where it should stay if at all possible.

The properties work like this:

On init (and after every state change via redux.subscribe), the user props are computed using the user-supplied `mapStateToComputed` function.  The resulting object might look like e.g.:

```js
// props
{
  users: [ { id: 1, name: 'Bob' } ]
}
```

The resulting props object is set on a private `_props` property on the component via `setProperties`, e.g.:

```js
this.get('_props').setProperties(props);
```

"read-only" aliases are set up to match what the user expects

```js
// Simplified
Ember.defineProperty(this, 'user', Ember.computed.readOnly('_props.user'));
```

This allows us to update the "private" component properties using redux.subscribe, and the user continues to consume the public read-only props (which are just aliases).  It is important to emphasize that _no additional state is being created_—just a pathway that allows us to make updates to the component and still keep everything locked down.

----

cc @brettburley

### DIFF
--- a/addon/components/connect.js
+++ b/addon/components/connect.js
@@ -1,6 +1,6 @@
 import Ember from 'ember';
 
-const { computed, defineProperty, run } = Ember;
+const { computed: { readOnly }, defineProperty, run } = Ember;
 
 var connect = function(mapStateToComputed, mapDispatchToActions) {
     var shouldSubscribe = Boolean(mapStateToComputed);
@@ -14,6 +14,7 @@ var connect = function(mapStateToComputed, mapDispatchToActions) {
             });
             return props;
         };
+
         var mapDispatch= function(dispatch) {
             var actions = [];
             Object.keys(finalMapDispatchToActions(dispatch)).forEach(function(key) {
@@ -21,6 +22,7 @@ var connect = function(mapStateToComputed, mapDispatchToActions) {
             });
             return actions;
         };
+
         return WrappedComponent.extend({
             redux: Ember.inject.service('redux'),
             init() {
@@ -29,38 +31,29 @@ var connect = function(mapStateToComputed, mapDispatchToActions) {
                 var redux = this.get('redux');
                 var props = mapState(redux.getState());
                 var dispatch = mapDispatch(redux.dispatch.bind(redux));
+                // Create a private object to hold private properties
+                this.set('_props', Ember.Object.create());
+                // Compute and set private properties based on current state
+                this.updateProps(redux.getState());
+                // Set public read-only aliases to the private properties
                 props.forEach(function(name) {
-                    defineProperty(component, name, computed(function() {
-                        return finalMapStateToComputed(redux.getState())[name];
-                    }).property().readOnly());
+                    defineProperty(component, name, readOnly(`_props.${name}`));
                 });
                 dispatch.forEach(function(action) {
                     component['actions'][action] = finalMapDispatchToActions(redux.dispatch.bind(redux))[action];
                 });
                 if (shouldSubscribe && !this.unsubscribe) {
-                    this.unsubscribe = redux.subscribe(this.handleChange.bind(this));
+                    this.unsubscribe = redux.subscribe(() => {
+                        run(() => {
+                            this.updateProps(redux.getState());
+                        });
+                    });
                 }
                 this._super(...arguments);
             },
-            handleChange() {
-                run(() => {
-                    var redux = this.get('redux');
-                    var props = mapState(redux.getState());
-                    var componentState = this.getComponentState(props);
-                    var reduxState = finalMapStateToComputed(redux.getState());
-                    props.forEach((name) => {
-                        if (componentState[name] !== reduxState[name]) {
-                            this.notifyPropertyChange(name);
-                        }
-                    });
-                });
-            },
-            getComponentState(props) {
-                var componentState = {};
-                props.forEach((name) => {
-                    componentState[name] = this.get(name);
-                });
-                return componentState;
+            updateProps(state) {
+                var props = finalMapStateToComputed(state);
+                this.get('_props').setProperties(props);
             },
             willDestroy() {
                 this._super(...arguments);


toranb:
@dustinfarris I like the simplicity of this for sure. One blocker I see currently is that setProperties is called with a new object each time (w/ Object.assign) likely resulting in a notifyPropertyChange for each property in the component. I can't say this is the case for certain so I'll need you and I to do some home work here.

Here is what I would like to accompany this PR regarding test coverage that I couldn't seem to find myself searching around for a few minutes tonight.

1. Throw out my monkey patch for `updateProps` and instead look deeper at what is truly invoked at the ember metal layer when notifyPropertyChange is fired.
2. Use this to get a new snapshot/metric before the changes you've set out here
3. Confirm your changes (and setProperties) doesn't incur any additional work from ember (as I suspect it does looking at setProperties here)

@brettburley if you have any working knowledge of ember internals (ie: what gets notified /when) it would greatly improve our ability to confirm we aren't back in over-notify land. I want to make sure we go above and beyond if we plan to refactor this because @justinpark opened an issue related to this very problem in #37 recently.

addon/components/connect.js
```diff
@@ -14,13 +14,38 @@ var connect = function(mapStateToComputed, mapDispatchToActions) {
...
+         * Transform properties state object into a private object
+         *
+         * Example:
+         *
+         *     transformPropsToPrivate({ users: [] })  // => { _users: [] }
+         */
+        var transformPropsToPrivate = function(props) {
+            // Transform object to new object with private keys
+            return Object.keys(props).reduce(function(privateState, key) {
+                return Object.assign(privateState, {
```
brettburley: Any advantage to using `Object.assign` here, rather than just assigning `privateState[getPrivateKey(key)] = props[key]`? I just ran a quick test in the console comparing the two and saw a ~50x difference in speed. Granted, I'm not sure how critical a few ms saved here is in the grand scheme of things.

addon/components/connect.js
```diff
@@ -14,13 +14,38 @@ var connect = function(mapStateToComputed, mapDispatchToActions) {
...
                 actions.push(key);
             });
             return actions;
         };
+
+        /**
+         * Transform a key into its private equivalent by prepending an underscore
+         */
+        var getPrivateKey = function(key) {
+            return `_${key}`;
```
brettburley: Any thoughts on storing the keys on the component itself vs. a private Ember.Object property like `_props` (ala ember-data's `_data` property)? In the latter, no key transformation is necessary which seems like a plus. Plus all the data is stored together (and eliminates issues where component users try to use an underscored version of a property for their own purposes and might conflict).

brettburley:
> One blocker I see currently is that setProperties is called with a new object each time (w/ Object.assign) likely resulting in a notifyPropertyChange for each property in the component.

Fortunately, `setProperties` will noop when the new value is equal to the current value. So it's safe to call it with an entirely new hash. We will see the issue we discovered here https://github.com/toranb/ember-redux/pull/30#issuecomment-258061853 where unchanged objects may be re-set as a result of the original being mutated by Ember at runtime. But the notification frequency should match the behavior with the current check here: https://github.com/toranb/ember-redux/blob/master/addon/components/connect.js#L52

Regarding tests, I think we should definitely be counting number of re-renders. One way to do that is like I did here by tracking `didRender`s: https://github.com/toranb/ember-redux/pull/30, but there might be other ways to count the same. It would probably be nice to add a few tests that only change properties that aren't rendered and verifying that the component isn't re-rendered.

We could also add some tests that set up observers on the properties and ensure the observers don't fire when the property doesn't change.

addon/components/connect.js
dustinfarris: Yeah, I like that!  I'll adjust and we can see how it looks.

addon/components/connect.js
dustinfarris: I don't think there's an advantage to `Object.assign`.  I'll make the change.

toranb:
@dustinfarris you mind implementing the re-render checks that @brettburley mentioned above? I was reading up on the [mixin](https://github.com/emberjs/ember.js/blob/v2.9.1/packages/ember-runtime/lib/mixins/observable.js#L30) that does observer work but I think re-render maybe the most appropriate give this approach. The biggest challenge I have with re-render only is that if I have 2 properties on the component and only 1 changes - I assume a re-render is invoked but did we (or did we not) notifyPropertyChange for both props?

toranb:
@dustinfarris could you push up the object.assign changes @brettburley mentioned? Without them we do indeed see a regression related to the notifyPropertyChange (using my new test in PR #42 )

here is the diff after using setProperties as you have it above

screen shot 2016-11-11 at 8 13 09 pm

brettburley:
I don't think `Object.assign` will change the behavior at all. Let's see after #42 is updated if the counts still mismatch... I would expect them to be the same in this case.

toranb:
@dustinfarris could you rebase this now that PR #42 have merged in? I found that your work w/ setProperties here is now regression free because ember already do the heavy lifting before it decides to notifyPropertyChange

screen shot 2016-11-19 at 3 19 25 pm

dustinfarris:
Rebased

dustinfarris:
I am addressing Brett's comments now

dustinfarris:
Ok, rebased and updated with @brettburley's suggestions—which simplified the code greatly as it is no longer necessary to transform the props object at all.

I've updated the description to match.  This should be good to go.

toranb:
@dustinfarris this is awesome! Just look at all the code we can delete! Thanks for all the hard work here

@brettburley this seems rock solid and ready to merge from my view. You like?

brettburley:
This looks great! Thanks for all the work @dustinfarris.

toranb:
@dustinfarris @brettburley this is now up on npm as part of the v1.9.0 release

PR #42, opened 2016-11-12
[TESTS]: updated rerender test to track notify via observer

toranb:
I've updated the rerender test to track each notifyPropertyChange more accurately. This should allow us to better evaluate the setProperties PR that Dustin has started on.

@dustinfarris you mind rebasing this after we merge it to see how the setProperties work compares?

@brettburley the only question I have before we land this - in the test I lookup the list-one component but since 2 instances are created ... how can I identify each independently so it's more clear in the test?

### DIFF
--- a/addon/components/connect.js
+++ b/addon/components/connect.js
@@ -50,7 +50,7 @@ var connect = function(mapStateToComputed, mapDispatchToActions) {
                     var reduxState = finalMapStateToComputed(redux.getState());
                     props.forEach((name) => {
                         if (componentState[name] !== reduxState[name]) {
-                            this.updateProps(name);
+                            this.notifyPropertyChange(name);
                         }
                     });
                 });
@@ -61,9 +61,6 @@ var connect = function(mapStateToComputed, mapDispatchToActions) {
                     componentState[name] = this.get(name);
                 });
                 return componentState;
-            },
-            updateProps(name) {
-                this.notifyPropertyChange(name);
             },
             willDestroy() {
                 this._super(...arguments);
--- a/tests/acceptance/rerender-test.js
+++ b/tests/acceptance/rerender-test.js
@@ -50,11 +50,11 @@ test('should only rerender when connected component is listening for each state 
         var componentTwo = application.__container__.lookup('component:list-two');
         var componentFour = application.__container__.lookup('component:unrelated-one');
         var componentFive = application.__container__.lookup('component:random-one');
-        componentOne.addObserver('items', () => oneUpdated++);
-        componentOne.addObserver('fake', () => oneFakeUpdated++);
-        componentTwo.addObserver('items', () => twoUpdated++);
-        componentFour.addObserver('unrelated', () => fourUpdated++);
-        componentFive.addObserver('random', () => fiveUpdated++);
+        componentOne.addObserver('items', function() { this.get('items'); oneUpdated++; });
+        componentOne.addObserver('fake', function() { this.get('fake'); oneFakeUpdated++; });
+        componentTwo.addObserver('items', function() { this.get('items'); twoUpdated++; });
+        componentFour.addObserver('unrelated', function() { this.get('unrelated'); fourUpdated++; });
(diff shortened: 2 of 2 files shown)


toranb:
@brettburley curious on your feedback generally - using this did seem to help me track down a legit increase in the number of notifyPropertyChanges fired in PR #41

brettburley:
@toranb Just left a comment -- other than that, I think this looks great! I suspect that if you do make that change though, you'll see the number of times observers might be increased.

toranb:
screen shot 2016-11-15 at 3 35 36 am

@brettburley it's all driven from the template like a real app would - I wouldn't expect we should/or need to do an additional `this.get` as you mentioned. Please let me know if I'm incorrect/misunderstand the test case we have here :)

brettburley:
I ran a quick test, and added the get to each of the observers:
```
@@ -50,11 +50,11 @@ test('should only rerender when connected component is listening for each state
         var componentTwo = application.__container__.lookup('component:list-two');
         var componentFour = application.__container__.lookup('component:unrelated-one');
         var componentFive = application.__container__.lookup('component:random-one');
-        componentOne.addObserver('items', () => oneUpdated++);
-        componentOne.addObserver('fake', () => oneFakeUpdated++);
-        componentTwo.addObserver('items', () => twoUpdated++);
-        componentFour.addObserver('unrelated', () => fourUpdated++);
-        componentFive.addObserver('random', () => fiveUpdated++);
+        componentOne.addObserver('items', function() { this.get('items'); oneUpdated++; });
+        componentOne.addObserver('fake', function() { this.get('fake'); oneFakeUpdated++; });
+        componentTwo.addObserver('items', function() { this.get('items'); twoUpdated++; });
+        componentFour.addObserver('unrelated', function() { this.get('unrelated'); fourUpdated++; });
+        componentFive.addObserver('random', function() { this.get('random'); fiveUpdated++; });
     });
```

When I did that, I saw slightly different results:
```
@@ -142,9 +142,9 @@ test('should only rerender when connected component is listening for each state
         assert.equal(find('.list-item-three .fake-value').text(), 1);
         assert.notEqual(find('.unrelated-one').text(), '');
         assert.notEqual(find('.random-one').text(), '');
-        assert.equal(oneUpdated, 3);
+        assert.equal(oneUpdated, 4);
         assert.equal(oneFakeUpdated, 0);
-        assert.equal(twoUpdated, 3);
+        assert.equal(twoUpdated, 4);
         assert.equal(fourUpdated, 1);
         assert.equal(fiveUpdated, 1);
     });
@@ -164,9 +164,9 @@ test('should only rerender when connected component is listening for each state
         assert.equal(find('.list-item-three .fake-value').text(), 5);
         assert.notEqual(find('.unrelated-one').text(), '');
         assert.notEqual(find('.random-one').text(), '');
-        assert.equal(oneUpdated, 3);
-        assert.equal(oneFakeUpdated, 2);
-        assert.equal(twoUpdated, 3);
+        assert.equal(oneUpdated, 4);
+        assert.equal(oneFakeUpdated, 4);
+        assert.equal(twoUpdated, 4);
         assert.equal(fourUpdated, 1);
         assert.equal(fiveUpdated, 1);
     });
```

I wonder if it's a result of the value being changed multiple times before a rendering? Given https://guides.emberjs.com/v2.9.0/object-model/observers/#toc_unconsumed-computed-properties-do-not-trigger-observers, I would recommend including the `get` in each observer to make sure that we are actually tracking all notifications of change to the properties. Otherwise, I think this is good to go!

toranb:
@brettburley I've included the changes you asked for in the latest commit. Still trying to wrap my head around the observer details you mentioned. But taking it on faith to see how the setProperties PR would respond after this lands I started stepping through the new connect that @dustinfarris wrote + this test and found what I was looking for 

screen shot 2016-11-19 at 3 19 25 pm

I now see that we don't over notify using the setProperties because ember already checks previous value == new value when deciding if it should notify.

That said, the setProperties PR does not introduce a regression now and I've personally step'd thru the code enough to know that's correct :)

Thanks for all the help on this one

PR #43, opened 2016-11-12
Update both ember and ember-cli to v2.9.1

toranb:
Keeping up with the cool kids - the only huge jump I see is from `ember-cli-qunit v2.1.0 => v3`


toranb:
Ah, had my previous commits on it - dumping this for a standalone ember-cli update PR

PR #44, opened 2016-11-12
Update both ember and ember-cli to v2.9.1 

toranb:
Keeping up with the cool kids - the only huge jump I see is from `ember-cli-qunit v2.1.0 => v3`

This time it's just the bower/package.json/ember-cli updates :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,5 +13,5 @@
 /connect.lock
 /coverage/*
 /libpeerconnection.log
-npm-debug.log
+npm-debug.log*
 testem.log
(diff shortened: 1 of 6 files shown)


toranb:
Interesting - suddenly after this update a handful of the first acceptance tests blow up in ember 2 and ember 2.4

toranb:
looks like all that hacking of the array prototype w/ replace got the best of me :)

`ret.replace is not a function`

toranb:
force pushed up a quick fix using splice -all is good now on both ember 2.4 and 2.9

dustinfarris:
LGTM

PR #46, opened 2016-11-20
Refactor connect

dustinfarris:
This is a **BREAKING CHANGE**

Simplifies `connect` by leveraging ES6 syntax and accepting `Component` as a primary argument.  This usage is more straight-forward and is easier to reason about internally.

Usage becomes, e.g.:

```js
const stateToComputed = (state) => {
  return { all: state.all };
}

const dispatchToActions = (dispatch) => {
  return { more: dispatch({ type: 'MORE' }) };
}

const MyComponent = Ember.Component.extend();

export default connect(MyComponent, stateToComputed, dispatchToActions);
```

### DIFF
--- a/addon/components/connect.js
+++ b/addon/components/connect.js
@@ -1,69 +1,103 @@
 import Ember from 'ember';
 
-const { computed: { readOnly }, defineProperty, run } = Ember;
+const {
+  computed: { readOnly },
+  defineProperty,
+  inject: { service },
+  isEmpty,
+  run
+} = Ember;
 
-var connect = function(mapStateToComputed, mapDispatchToActions) {
-  var shouldSubscribe = Boolean(mapStateToComputed);
-  var finalMapStateToComputed = mapStateToComputed || function() {return {};};
-  var finalMapDispatchToActions = mapDispatchToActions || function() {return {};};
-  return function wrapWithConnect(WrappedComponent) {
-    var mapState = function(state) {
-      var props = [];
-      Object.keys(finalMapStateToComputed(state)).forEach(function(key) {
-        props.push(key);
-      });
-      return props;
-    };
+/**
+ * Modifies a Component by mapping redux state and dispatch to its properties
+ * and actions respectively.
+ *
+ * ```javascript
+ * const stateToComputed = (state) => {
+ *   return { all: state.all };
+ * }
+ * const dispatchToActions = (dispatch) => {
+ *   return { more: dispatch({ type: 'MORE' }) };
+ * }
+ * const MyComponent = Ember.Component.extend();
+ *
+ * export default connect(MyComponent, stateToComputed, dispatchToActions);
+ * ```
+ *
+ * @function connect
+ * @param {Ember.Component} Component the Ember component to be connected
+ * @param {function} stateToComputed A function that maps redux state to an
+ *   object of desired properties that will be added to the component
+ * @param {function} dispatchToActions A function that maps component
+ *   actions to redux dispatch
+ * @returns {Ember.Component} The original component with mapped properties
+ *   and actions
+ */
+export default function connect(Component, stateToComputed = () => ({}), dispatchToActions = () => ({})) {
 
-    var mapDispatch= function(dispatch) {
-      var actions = [];
-      Object.keys(finalMapDispatchToActions(dispatch)).forEach(function(key) {
-        actions.push(key);
-      });
-      return actions;
-    };
+  return Component.extend({
 
-    return WrappedComponent.extend({
-      redux: Ember.inject.service('redux'),
-      init() {
-        var component = this;
-        component['actions'] = Ember.$.extend({}, component['actions']);
-        var redux = this.get('redux');
-        var props = mapState(redux.getState());
-        var dispatch = mapDispatch(redux.dispatch.bind(redux));
-        // Create a private object to hold private properties
-        this.set('_props', Ember.Object.create());
-        // Compute and set private properties based on current state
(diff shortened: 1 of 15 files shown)


toranb:
Breaking change and it differs from the react-redux api ... not in favor personally as discussed in the open issue

toranb:
@dustinfarris after my initial reaction above here is why I will be closing this out

1) no breaking api changes in 2016

As we discussed several times I'm not planning a 2x release in 2016 to allow consumers of the addon to build on it /learn/ etc. You may not be aware but I've got a written guide series, screen cast series and source code in flight for my guides/training/intro overhaul and releasing those w/ an already out of date api is poor planning.

2) the connect api is the heart of this addon and I hold firm that I want to keep it in sync with the api of react-redux. Making a shift like this fractures the api for people (like me) who want a consistent redux api for components in ember/react/angular.

From what I can tell you disagree and that's cool - if you feel strongly enough you can fork this and do whatever you please (the power of open source)

3) The final word - making this change may simplify the internals for you, but I find the connect internals as they now to be straight forward. This debate has been about preference since it's inception and it's time I made myself clear. I'm not interested in a refactor/ or change of any kind to the connect api for components.

I'm at the point now that I have anxiety when you submit an issue or PR because I know you hate the api and prefer something different. I would like it if I didn't feel this way - I hope you understand my firm "no, but thank you" response here tonight because it has given me some sense of peace.

dustinfarris:
1. This didn't have to be merged immediately — I get that you have a big presentation coming up and I apologize for not being up front with that consideration
2. I don't know enough about react-redux/angular whatever, so I'd ask for a little more leniency here — the breaking change is actually a very small part of this PR which could be easily reversed with perhaps some discussion first since i am ignorant
3. For the most part, I believe that I've done well here with regard to the simplification/refactoring.  We discussed that I would submit a refactor proposal and I spent a good part of yesterday thinking about how that would work and the better part of today piecing it together

With regards to the breaking part of this PR, I don't "hate the API", I just find that it would make more sense this way.  I fully understand that that's subjective.  It really makes no difference to me in the end—it works just the same which is all I truly care about.

You shouldn't have anxiety over my contributions.  We are both reasonable adults, and I understand that I don't have the best ideas all the time.  However, I would appreciate being able to respond, here or elsewhere, before my work is dismissed.

dustinfarris:
@toranb here it is with the API change reverted (and docstrings stripped so the code is clear):

https://github.com/dustinfarris/ember-redux/blob/es6/addon/components/connect.js

I humbly request that you give this a chance.  I understand your position on the API and I'm fine with it, but I do believe I have added value by simplifying the internal logic, which was the core purpose of this refactor even though it didn't come across that way.

dbashford:
Should note, this is also a breaking change in that it leverages `Object.assign`.  Not a huge deal, but previously no polyfills were necessary.

We, rather than use `Object.assign`, instead use the spread operator. For example:

```javascript
  [ACTION_TYPES.SET_EVENT_TYPE]: (state, { payload: eventType }) => ({
    ...state,
    eventType
  }),
```

Object spread can be transpiled. `Object.assign` has to be polyfilled. I was a little surprised to see that we hadn't include the babel polyfill, and hadn't yet needed it. We can include it. But just in case anyone sees an issue, felt I would mention it. =)

dustinfarris:
Great point!  Doesn't object spread require the ES7 option though?  Also I don't think Object.assign itself needs the polyfill, but the variable key name does.  Possible issue for older node and maybe PhantomJS which has finicky support for this.

dbashford:
Ya, ES7. But I've longed stopped tracking what ES version things belong to. It is about what babel can do for me. Turn it all on! Point being, it is transpiled so we hadn't noticed that we hadn't included the polyfill. And even then we didn't notice the problem until we ran our test suite. Phantom blew up. 

One line update to our config and we are good, but figured I'd toss it in here in case anyone runs into the same problem.

toranb:
@dbashford this is unfortunate :( and the only reason it broke you and not us is that we run tests w/ firefox and chrome (not phantomJS)

What polyfill did you add in particular to get your test suite running? Is this just a 1 liner to tell babel to include it?

dbashford:
Yep, just a [one liner in the ember-cli-babel config](https://github.com/babel/ember-cli-babel#polyfill).

dustinfarris:
We could use the [Ember polyfill](http://emberjs.com/api/#method_assign) for Object.assign perhaps

toranb:
@dbashford I feel like the most correct thing here is to add the polyfill for people using this addon as @dustinfarris mentioned (if they haven't already) by adding this include in the ember-cli-build of ember-redux. I assume that is only included 1x and not a burden if the parent app already has the flag set.

Any downsides to altering this? It seems to fix the semver break and it's likely a good idea to prevent a simple error from failing a build for someone new to redux in ember. thoughts?

dustinfarris:
@toranb the ember-cli-build in the addon only affects the build of the dummy app when you test/serve it while developing the addon itself.

Our options are to either

1. Polyfill our own build in treeForApp by running the code through babel with the polyfill flag on
2. Ask users to do this in their own ember-cli-build on an as-needed basis
3. Adjust the addon code to not need polyfill for example by using Ember.assign instead of Object.assign

I would prefer the second or third option myself

dustinfarris:
*First or third

toranb:
@dustinfarris ah, good call. About Ember.assign ... because it's only ember 2.5+ I don't like that option (trying to keep this ember 2.0+ friendly). With the choices I prefer to leave it as-is and ask the user to polyfill as needed

PR #49, opened 2016-11-21
Namespace private container name

dustinfarris:
"_props" seems fairly generic and might be used by other addons/libraries.  Renaming this to "_reduxProps" is more specific to what we are doing—less likely to be clobbered.

### DIFF
--- a/addon/components/connect.js
+++ b/addon/components/connect.js
@@ -2,6 +2,8 @@ import Ember from 'ember';
 
 const { computed: { readOnly }, defineProperty, run } = Ember;
 
+const PRIVATE_CONTAINER = '_reduxProps';
+
 var connect = function(mapStateToComputed, mapDispatchToActions) {
   var shouldSubscribe = Boolean(mapStateToComputed);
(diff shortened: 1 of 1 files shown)


toranb:
@dustinfarris thanks for the quick PR but I honestly want to patch this quick before taking on any additional changes (to reduce risk/ know for certain what is needed and why). I want to work with @dbashford to better understand what exactly he is doing and why we didn't see this locally (so I can construct something in our local test suite to catch regressions in the future).

The other reason I don't want to patch anything else it is that I plan to revert the known PR once @dbashford confirms it so others who pick this up don't run into an immediate issue and punt.

dustinfarris:
@toranb ok, i'd like @dbashford to attempt this PR quickly before we revert everything.  he has agreed to do this

toranb:
@dustinfarris sounds like a plan -thanks for the follow up

toranb:
@dustinfarris closing this out until after @dbashford has time to determine root cause (also because of the revert this code won't be merge friendly)

PR #50, opened 2016-11-25
[REVERT]: undo the setProperties update (from PR #41)

toranb:
At this time we have a known regression that community member @dbashford found after the v1.9 release. I had him try the connect we had prior to the setProperties pull request #41 and he confirmed that before that change all was working as it did in the prior v1.8. Until we can confirm the exact root cause, I feel the project is at risk and could ultimately turn people away if they run into this same problem while we work through this regression /and patch the addon to solve whatever incompatibility we discover.

The only downside is that we will need to put other connect refactor pull requests on hold while we work closely with @dbashford to reproduce this.

### DIFF
--- a/addon/components/connect.js
+++ b/addon/components/connect.js
@@ -1,6 +1,6 @@
 import Ember from 'ember';
 
-const { computed: { readOnly }, defineProperty, run } = Ember;
+const { computed, defineProperty, run } = Ember;
 
 var connect = function(mapStateToComputed, mapDispatchToActions) {
   var shouldSubscribe = Boolean(mapStateToComputed);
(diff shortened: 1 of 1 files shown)


dbashford:
Apologize for not being able to get on this straight away. Spent the last few days working on run loop testing stability issues, and now Thanksgiving.

I suspect I have another day of testing work to do.

toranb:
@dbashford no problem - I'm off for a long holiday weekend myself. I wanted to revert so new developers who might pick this up over the break have a stable/solid version to work with. I'm certain we will find out why this didn't work but to avoid rushing you (and keep the addon truly stable) I think it's best we revert until we know more about this corner case.

PR #47, opened 2016-11-21
Refactor connect implementation

dustinfarris:
Clear up the internals of connect:

- Use es6 syntax
- Refactor some logic
- Docstrings

PR #52, opened 2016-11-29
ES6ify connect

dustinfarris:
This replaces #47 

No functional changes, just syntax sugar and move `finalMapToComputed`/`finalDispatchToActions` to default args.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -14,6 +14,7 @@
   "forin": false,
   "immed": false,
   "laxbreak": false,
+  "loopfunc": true,
   "newcap": true,
   "noarg": true,
   "noempty": false,
--- a/addon/components/connect.js
+++ b/addon/components/connect.js
@@ -1,58 +1,57 @@
 import Ember from 'ember';
 
-const { computed, defineProperty, run } = Ember;
+const {
+  computed,
+  defineProperty,
+  inject: { service },
+  isEmpty,
+  run
+} = Ember;
 
-var connect = function(mapStateToComputed, mapDispatchToActions) {
-  var shouldSubscribe = Boolean(mapStateToComputed);
-  var finalMapStateToComputed = mapStateToComputed || function() {return {};};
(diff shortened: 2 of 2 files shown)


dustinfarris:
Also, handleChange is moved to an anonymous function.  I can undo that part if `handleChange` is considered public API.

.jshintrc
```diff
@@ -14,6 +14,7 @@
   "forin": false,
   "immed": false,
   "laxbreak": false,
+  "loopfunc": true,
```
toranb: @dustinfarris is this loopfunc still required?

addon/components/connect.js
```diff
@@ -1,58 +1,57 @@
...
-        dispatch.forEach(function(action) {
-          component['actions'][action] = finalMapDispatchToActions(redux.dispatch.bind(redux))[action];
-        });
-        if (shouldSubscribe && !this.unsubscribe) {
-          this.unsubscribe = redux.subscribe(this.handleChange.bind(this));
+        const redux = this.get('redux');
+
+        let props = stateToComputed(redux.getState());
+
+        for (let name in props) {
```
toranb: why favor the for /in over Object.keys here?

.jshintrc
dustinfarris: Yes, for line 25

toranb: also noticed we don't have a "return" keyword here anymore - is that implicit now?

addon/components/connect.js
toranb: Looking around the web it appears Object.keys is more performant and how this works today. Let's keep these enumerations as Object.keys

dustinfarris: Ok, using Object.keys and rebased

.jshintrc
toranb: Oh I mean in the computed itself - we had an explit return previously

Issue #54, opened 2016-11-30
Add "ownProps" argument for stateToComputed/dispatchToActions


dustinfarris:
This is something react-redux does.  For us it might look like:

```hbs
{{user-detail userId=567}}
```

```js
const stateToComputed = (state, ownProps) => {
  return {
    user: getUserById(state, ownProps.userId)
  }
}
```

I think this should be simple to add.  react-redux defines this argument as

> the props passed to your component

so we're not talking about all possible properties on the Ember Component, just what was passed to it, which Ember stores in `attrs`.

We can pass in attrs to the functions, and then listen to [didUpdateAttrs](https://github.com/emberjs/ember.js/blob/v2.10.0/packages/ember-glimmer/lib/component.js#L354)

so connect would be something like

```js

...

handleChange() {  // and also in init
  ...
  let props = stateToComputed(redux.getState(), this.attrs);
  ...
},

didUpdateAttrs() {
  this.handleChange();
}
```

dustinfarris:
Unless someone sees something wrong with this, I will spin up a PR tomorrow.

wooandoo:
Hello,

I suggest:
```js
const attrValues = (attrs) => Object.keys(attrs).reduce((result, key) => {result[key] = attrs[key].value; return result}, {});

// ...

let props = stateToComputed(redux.getState(), attrValues(this.attrs));

didUpdateAttrs() {
  this._super(...arguments);
  this.handleChange();
},
```

This avoids to get "value" for me:
```js
ownProps.userId.value
```

Frédéric

dustinfarris:
Great suggestion! 👍 thanks!

toranb:
@dustinfarris can we close this issue out and drive the discussion from that open PR ?

dustinfarris:
@toranb I put a close hook in the commit message so this should close automatically once merged to master—not sure what organizational policy you prefer in terms of managing open issues/prs.  Maybe we should discuss offline someday to get on same page.

toranb:
@dustinfarris ah right on :)

PR #56, opened 2016-12-01
[FEATURE] pass attrs to stateToComputed

dustinfarris:
### Summary

This adds a second argument to `stateToComputed` that contains an object with the attrs that were passed to a component.

E.g.:

```hbs
{{my-component userId=userId}}
```

```js
stateToComputed = (state, attrs) => {
  return { user: getUserById(state, attrs.userId) };
}
```

This is analogous to [the same API in react-redux](https://github.com/reactjs/react-redux/blob/master/docs/api.md#connectmapstatetoprops-mapdispatchtoprops-mergeprops-options).

### Details

As suggested by @wooandoo in #54, rather than pass `this.attrs` and leave it at that, a new object is computed that alleviates the need to crawl through the intermediary attr.  This object is cached for performance and only updated if the inbound attrs change as signaled by `didUpdateAttrs`.  More details can be found in the included docstring for `computeAndCacheAttrs()`.

### Also

react-redux does the same thing for `mapDispatchToProps`, however this is not so easy in Ember.  It seems that after actions are defined, they cannot be easily redefined—therefore the action would not update if the attrs changed, as the user would expect.  This PR leaves `dispatchToActions` alone until this can be investigated further.

### Alternative

Rather than putting together an attrs object, we could just pass the entire component itself, e.g.:

```js
// connect.js
...
let props = stateToComputed(redux.getState(), this);
```

```hbs
{{my-component userId=userId}}
```

```js
stateToComputed = (state, component) => {
  return { user: getUserById(state, component.get('userId')) };
}
```

This would provide a lot more flexibility and simplify the internal implementation, however it would be a little out of step with react-redux, and opens the door for possible misuse.

/cc @toranb

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,10 +31,11 @@ import Ember from 'ember';
 import hbs from 'htmlbars-inline-precompile';
 import connect from 'ember-redux/components/connect';
 import ajax from 'example/utilities/ajax';
+import getUsersByAccountId from '../reducers';
 
-var stateToComputed = (state) => {
+var stateToComputed = (state, attrs) => {
   return {
-    users: state.users.all
+    users: getUsersByAccountId(state, attrs.accountId)
   };
 };
 
@@ -73,7 +74,7 @@ export default UserTableComponent;
 
 ## Example Composition
 ```js
-{{#user-list as |users remove|}}
+{{#user-list accountId=accountId as |users remove|}}
   {{user-table users=users remove=remove}}
 {{/user-list}}
 ```
--- a/addon/components/connect.js
+++ b/addon/components/connect.js
@@ -19,11 +19,11 @@ export default (stateToComputed=() => ({}), dispatchToActions=() => ({})) => {
       init() {
         const redux = this.get('redux');
 
-        let props = stateToComputed(redux.getState());
+        let props = stateToComputed(redux.getState(), this.getAttrs());
 
         Object.keys(props).forEach(name => {
           defineProperty(this, name, computed(() =>
-            stateToComputed(redux.getState())[name]
+            stateToComputed(redux.getState(), this.getAttrs())[name]
           ).property().readOnly());
         });
 
@@ -43,13 +43,40 @@ export default (stateToComputed=() => ({}), dispatchToActions=() => ({})) => {
       handleChange() {
         const redux = this.get('redux');
 
-        let props = stateToComputed(redux.getState());
+        let props = stateToComputed(redux.getState(), this.getAttrs());
 
         Object.keys(props).forEach(name => {
           if (this.get(name) !== props[name]) {
             this.notifyPropertyChange(name);
           }
         });
+      },
+
+      /**
+       * Return an object of attrs passed to this Component.
+       *
+       * `Component.attrs` is an object that can look like this:
+       *
+       *   {
+       *     myAttr: {
+       *       value: 'myValue'
+       *     }
+       *   }
+       *
+       * Ember provides that a `get` will return the value:
+       *
+       *   this.get('myAttr') === 'myValue'
+       *
+       * @method getAttrs
+       * @private
+       */
+      getAttrs() {
+        return this.getProperties(Object.keys(this.attrs || {}));
+      },
+
+      didUpdateAttrs() {
+        this._super(...arguments);
+        this.handleChange();
       },
 
       willDestroy() {
--- a/tests/integration/connect-test.js
+++ b/tests/integration/connect-test.js
@@ -38,6 +38,18 @@ test('should render parent component with one state and child component with ano
   assert.equal($child.text(), 7);
 });
 
+test('should render attrs', function(assert) {
+  assert.expect(2);
+  this.set('myName', 'Dustin');
+  this.render(hbs`{{count-list name=myName}}`);
+
+  assert.equal(this.$('.greeting').text(), 'Welcome back, Dustin!', 'should render attrs provided to component');
+
+  this.set('myName', 'Toran');
+
+  assert.equal(this.$('.greeting').text(), 'Welcome back, Toran!', 'should rerender component if attrs change');
+});
+
 test('the component should truly be extended meaning actions map over as you would expect', function(assert) {
   this.render(hbs`{{count-list}}`);
   let $random = this.$('.random-state');
@@ -59,10 +71,15 @@ test('each computed is truly readonly', function(assert) {
 });
 
 test('lifecycle hooks are still invoked', function(assert) {
-  assert.expect(2);
+  assert.expect(3);
   this.register('component:test-component', connect()(Ember.Component.extend({
     init() {
       assert.ok(true, 'init is invoked');
+      this._super(...arguments);
+    },
+
+    didUpdateAttrs() {
+      assert.ok(true, 'didUpdateAttrs should be invoked');
       this._super(...arguments);
     },
 
@@ -72,5 +89,7 @@ test('lifecycle hooks are still invoked', function(assert) {
     }
   })));
 
-  this.render(hbs`{{test-component}}`);
+  this.render(hbs`{{test-component name=name}}`);
+
+  this.set('name', 'Dustin');
 });
--- a/tests/dummy/app/components/count-list/component.js
+++ b/tests/dummy/app/components/count-list/component.js
@@ -2,10 +2,11 @@ import Ember from 'ember';
 import hbs from 'htmlbars-inline-precompile';
 import connect from 'ember-redux/components/connect';
 
-var stateToComputed = (state) => {
+var stateToComputed = (state, attrs) => {
   return {
     low: state.low,
-    high: state.high
+    high: state.high,
+    greeting: `Welcome back, ${attrs.name}!`
   };
 };
 
@@ -24,6 +25,7 @@ var CountListComponent = Ember.Component.extend({
     <button class="btn-random" onclick={{action "random"}}>random</button>
     <button class="btn-alter" onclick={{action "alter"}}>alter</button>
     <span class="random-state">{{color}}</span>
+    <span class="greeting">{{greeting}}</span>
   `,
   actions: {
     alter() {


dustinfarris:
Updated summary with alternative section

foxnewsnetwork:
lol I was just looking up how to get component attributes from into `stateToCompute` and, imo as a humble end-user, this is definitely something that we totally need, so 👍 

Although, I would also suggest yet another alternative where ember-redux just ships both the `attrs` version as well the `this` version; i.e.:

`ember-redux/addons/connect.js`

```javascript
let props = stateToComputed(redux.getState(), this._cachedAttrs, this);
```
and in the docs, we just don't openly mention we're passing in `this` as the third argument - let it be a sort of naughty secret that only truly desperate users uses.

Why this is a good idea is because a lot of us casuals are stuck with legacy code that must fetch data out of injected services or whatnot that aren't passed in on attributes. Trying to get an IRL org to switch over to the redux paradigm is already enough of a battle, so, imho, sometimes it's necessarily to have this sort of escape hatch to allow us to pull stuff out of `inject.service` or whatever... After all, management apparently really doesn't like funding ideas that require putting everything on-hold for several months for a re-write.

wooandoo:
@foxnewsnetwork I'm agree.
But it's even more true for dispatch/action to avoid to use an "old" function instead of an arrow function.

For the properties, you can call your service in the route and inject the "state property" in the component as attributes, and so, well define your immutable and "pure" state. ;-)

Frédéric

toranb:
@foxnewsnetwork I planned to (officially) document the implicit `this` argument that comes along with `stateToComputed` when I release the new guides. I assume if we include attrs like @dustinfarris has implemented here we would still get `this` as the last argument (correct me if I'm wrong).

That said, this raises an interesting question about semver/ public and intimate apis. If we landed this feature in 1x would it be a "breaking change" ? Assuming 2 use cases for the moment

1. I only ever used the happy path/ public api of `stateToComputed` making this PR additive 

2. I had to work around a hole in the api and started using the unofficially supported `this` param to pull in attrs/or other local properties found on the ember component making this PR something that would break my app (as the 2nd argument is suddenly not `this` and we bump `this` to the 3rd argument)

@dbashford @justinpark @brettburley @dustinfarris @foxnewsnetwork @wooandoo any thoughts on this PR breaking backwards compatibility? For a little more context, we are planning a 2x release anyway because v2+ will drop the ember-browserify requirement we have today (and this requires ember-cli v2.9.1+). This should speed up both hot and cold rebuild times for everyone using this addon (assuming they've only added browserify for this addon specifically).

I'm personally leaning toward "it's a breaking change" myself and something we will only support in the 2x branch as it would break any app in production today that makes an assumption of `this` as the 2nd argument.

README.md
```diff
@@ -31,10 +31,11 @@ import Ember from 'ember';
 import hbs from 'htmlbars-inline-precompile';
 import connect from 'ember-redux/components/connect';
 import ajax from 'example/utilities/ajax';
+import getUsersByAccountId from '../reducers';
```
toranb: This is a great simplistic use case! I'm wrestling with the idea of our most basic example showing this off vs adding this to the guides/gh-pages documentation (and in doing so, actually adding a new page for the connect api like you see in react-redux) -thoughts on the pros/cons of either decision ?

addon/components/connect.js
```diff
@@ -52,6 +54,44 @@ export default (stateToComputed=() => ({}), dispatchToActions=() => ({})) => {
...
+       * @private
+       */
+      computeAndCacheAttrs() {
+        this._cachedAttrs = Object.keys(this.attrs || {}).reduce((attrs, key) => {
+          return Object.assign(attrs, { [key]: this.get(key) });
+        }, {});
+      },
+
+      didUpdateAttrs() {
+        this.computeAndCacheAttrs();
```
toranb: can you add a test for the missing super call? look for the last merged PR from Brett for the simplest 

`this._super(...arguments);`

addon/components/connect.js
```diff
@@ -52,6 +54,44 @@ export default (stateToComputed=() => ({}), dispatchToActions=() => ({})) => {
...
+       *
+       *   this.computeAndCacheAttrs();
+       *   console.log(this._cachedAttrs);  // Object { myAttr: 'myValue' }
+       *
+       * @method computeAndCacheAttrs
+       * @private
+       */
+      computeAndCacheAttrs() {
+        this._cachedAttrs = Object.keys(this.attrs || {}).reduce((attrs, key) => {
+          return Object.assign(attrs, { [key]: this.get(key) });
```
toranb: this `Object.assign` in place to protect us from people changing the attrs? I'm curious if we should instead `Object.freeze` this so users would get an error

thoughts? I'd need to research the implications of this - maybe we could just freeze in debug mode like ember is doing now with the helpers param args?

tests/integration/connect-test.js
```diff
@@ -38,6 +38,18 @@ test('should render parent component with one state and child component with ano
...
+test('should render attrs', function(assert) {
+  assert.expect(2);
+  this.set('myName', 'Dustin');
+  this.render(hbs`{{count-list name=myName}}`);
+
+  assert.equal(this.$('.greeting').text(), 'Welcome back, Dustin!', 'should render attrs provided to component');
+
+  this.set('myName', 'Toran');
+
+  assert.equal(this.$('.greeting').text(), 'Welcome back, Toran!', 'should rerender component if attrs change');
```
toranb: love the simplicity of this test case - great work on this pull request Dustin!!!! Truly grateful to have someone like you contributing to this project!

dbashford:
All sounds good to me. We're upgrading to 2.10 this week, probably, so we should be good with whatever changes come. Removing the browserify dependency will be nice, and the addition of the attrs and `this` will also come in handy.

README.md
dustinfarris: Oh yeah, I'm all for more docs and keeping the README basic.  I'm thinking the improved documentation can be a separate PR—at which point this can be removed from the README.

On that note, we should probably have an "API" page of the docs, but that's a separate issue.

addon/components/connect.js
dustinfarris: The intention of this method is to simplify the `attrs` object

From this:

```js
{
  myAttr: {
    value: "myValue"
  }
}
```

to this:

```js
{
  myAttr: "myValue"
}
```

Disconnecting the 2-way binding is just a welcome side-effect of doing that.  I'm open to Object.freeze to make that explicit but do not feel strongly one way or the other.  Let me know if you want this.

toranb: @dustinfarris ah, in that case ignore my mention of freeze :)

I remember reading this comment in the original issue - looking forward to it !

README.md
toranb: @dustinfarris agreed on the API page you suggest :)

toranb:
@dustinfarris do you have time this week to update this PR so we can look at merging it sometime soon?

dustinfarris:
Yes I'll wrap it up tomorrow

addon/components/connect.js
brettburley: Small nit, but I think you could also do:
```
const keys = Object.keys(this.attrs || {});
this._cachedAttrs = this.getProperties(keys);
```

dustinfarris: @brettburley great point! i wonder if we even need to cache the result at that point:

e.g. if we do

```js
stateToComputed(redux.getState(), this.getProperties(Object.keys(this.attrs || {})));
```

brettburley:
This looks great to me 👍 

Regarding your alternative, I do like the API of getting the component directly, which seems more "ember-like" to me, since you could do things like `get('someProperty')`, but obscures when the `stateToComputed` method is called. It makes sense to be called whenever there's new state, or a new attr, so directly receiving the attrs as you have written seems to be the safest option; there's no concern that you might be using something in the function that wouldn't trigger it being recomputed.

>  I planned to (officially) document the implicit this argument that comes along with stateToComputed when I release the new guides.

@toranb is this the actual current behavior? I don't immediately see why this would be the case, and couldn't reproduce in the dummy app in the tests. Is the second argument passed into `stateToComputed` the component instance?

toranb:
@brettburley I'll see if I can't put together an example of when I was able to use this. Off the top of my head I'm now thinking it was `this` in context of dispatch (using the arrow function maybe?) that allowed me to get at the component properties /etc.

@foxnewsnetwork how are you getting access to `this` today? Just curious if you could clarify that "yes I have this/ here is how I pulled it in" or "nope- wish list item dude"

addon/components/connect.js
dustinfarris: Done

dustinfarris:
- Internal implementation has been greatly simplified thanks to suggestion from @brettburley to use getProperties instead of reduce.
- Added a test to ensure didUpdateAttrs calls super.

After this is merged we will need a follow up issue/PR to make the documentation changes that were discussed.

foxnewsnetwork:
@toranb TL;DR: it's on my wish-list lol.

#### Longer Answer 
Currently, I'm working around having to access `this` by doing a lot of stuff in the `.hbs` files. For example, consider my latest use case where I'm building some sort of anime catalog service. Say I have a `animes/anime/:anime_id` route and a `routes/anime.js` route. In the catalog route, I have something like:

```javascript
function model(dispatch) { 
  return dispatch(createAction(LIST_ANIMES));
}
```
Then, in my `routes/animes/anime.js` file, I would have:

```javascript
function model(dispatch, { anime_id }) { 
  return anime_id;
}
```
Then, in my `templates/animes/anime.hbs` file, I have:

```handlebars
{{#delegate-anime-detail slug=model user as |anime|}}
  {{#anime-detail-overview anime=anime as |sections|}}
    {{#if sections.isHeader}}
      <!-- header stuff -->
    {{else if sections.isCTA}}
      <!-- cta stuff -->
    {{else if ...}}
      ...
    {{/if}}
  {{/anime-detail-overview}}
{{/delegate-find-anime}}
```

Finally, (here's the non-ideal part), in my `components/delegate-anime-detail.js` and `templates/components/delegate-anime-detail.hbs`, I have no choice but resort to using a `find-by` helper in the template file as I can't properly pass in attributes (or this) into the `statesToCompute` function:

```javascript
function statesToCompute(state) { 
  return { animes: getAnimes(state.orm) };
}
```

```handlebars
{{yield (find-by 'id' slug animes)}}
```

Having to fetch all my `animes` just to show a single one is obviously non-ideal, but because I'm using redux-orm and friends, performance actually isn't a problem... the real issue comes from the inconvenience of being forced to work around javascript in the hbs files which, while not a big deal right now when all I'm doing is finding by id, will become significantly more painful if I have to create something like a `user-session-query-by` helper just so I can:

```handlebars
{{yield (user-session-query-by someQueryParams remoteAnimesRef)}}
```

#### Musings
Actually, as yet another work-around for the can't-access-state problem in components, I was tempted to dispatch some sort of reference to the current routing scheme (say the `currentRouteName`, the associated dynamic-id-context stack, and the query-params) into my redux store, but I decided against doing this properly requires changes on the ember framework level and that's not something I'm comfortable handling on the application level. 

Furthermore, I wasn't comfortable with the idea of components (which were suppose to be usable anywhere) being coupled to the routing structure and thus being able to only live in certain route templates. It smelled too much of ye olde' `registerChild` anti-pattern where child components would register themselves with their parent components who would in turn register themselves with the `Ember.Controller` (except worse because instead of the buck stopping at the controller level, we now go all the way to the top to the global state level).

And having been bitten once (who am I kidding, hundreds of times on a daily basis) by that particular anti-pattern (which, incidentally, is definitely a solution pressured engineers can easily jump into given business pressure), I'm all for passing in `attrs` and even `this` into the second and third argument slots to `statesToCompute`

toranb:
@brettburley I just confirmed that I'm either crazy or a total liar :)

Using both legacy functions/ and arrow functions never produces `this` as I mentioned above. This is a good thing because we can add `this` as a 3rd argument for `stateToComputed` w/out fear of breaking anything in 1x

toranb:
@brettburley so I'm not truly crazy after all :) just looked one more time and realized w/ that the arrow function `this` is a thing but it's the module (not the component instance as I remembered)

screen shot 2016-12-06 at 6 47 24 pm

toranb:
@dustinfarris I can take on the `this` addition @foxnewsnetwork is asking for after we merge this pull request. What all is left to look at /improve here before we merge this?

toranb:
@dustinfarris the only thing not changed is the documentation around this. You planning to pull those and make a stand alone example for the gh-pages page? or did I misunderstand that exchange ?

dustinfarris:
Right, the addition to README here is a quick fix.  Will follow up with separate PR to address docs and adjust README as we discussed.

PR #58, opened 2016-12-06
[BUG] stateToComputed and Ember binding

dustinfarris:
## Summary

Passing objects from redux state to other components results in Ember binding which affects the state itself.  This affects the purity/immutability of redux, and has the potential be problematic in practice especially if the user is unaware that this is happening.

## Details

Let's say current redux.getState() looks like this:

```
redux.getState();

{
  project: {
    name: 'Chores'
  }
}
```

And you have a "container" component that looks like this:

```js
const stateToComputed = (state) => {
  return {
    project: state.project
  }
}

const ProjectContainerComponent = Component.extend({
  layout: hbs`{{project-detail project=project}}`
});
```

As soon as you pass the computed property, `project`, to another component, Ember "binds" it so it can watch for changes.  Because this is all passed down by reference, the transformation that Ember performs to accomplish this go all the way back to redux state!

Now, your redux state looks something like this (depending on which browser you're using to inspect):

```
redux.getState();

{
  project: {
    __ember_meta__: { ... }
    getter: ...
    setter: ...
    name: 'Chores'
  }
}
```

The failing test in this PR shows one example of how this can bite you:

- The project-container component invokes the project-presentation component and binds the `project` computed property as provided by stateToComputed
- The test attempts a dispatch that updates the project in redux state, `dispatch({ type: 'UPDATE_PROJECT', newData: { name: 'New Name' } })`
- The dispatch handler attempts `Object.assign(state, newData)` which results in an error

```
You must use Ember.set() to set the `name` property (of [object Object])
```

## Solution

In the computed, we could deep copy stateToComputed before peeling off the desired property:

```js
// connect.js

defineProperty(this, name, computed(() =>
  Ember.$.extend(true, {}, stateToComputed(redux.getState()))[name]
).property().readOnly());
```

## Alternative

Do nothing.  Even though Ember modifies the state in this way, it may be possible to work with it nonetheless.  For example, with Object.assign, prepending and empty object works:

```js
Object.assign({}, boundState, newObject);  // works
Object.assign(boundState, newObject);  // throws error
```

Users can also deep copy on their own on a case-by-case basis:

```js
stateToComputed = (state) => {
  return {
    project: Ember.$.extend(true, {}, state.project)
  };
};
```

This could potentially get very expensive, and the user would need to know about this nuance of Ember, and how to deal with it.  Documentation would probably be appropriate.

### DIFF
--- a/tests/acceptance/connect-e2e-test.js
+++ b/tests/acceptance/connect-e2e-test.js
@@ -70,3 +70,15 @@ test('components without state should not subscribe or unsubscribe', function(as
     assert.equal(unsubscribed, 1);
   });
 });
+
+
+test('binding computed props does not affect redux state', function(assert) {
+  visit('/project');
+  assert.deepEqual(redux.getState().project, { name: 'Chores' });
+
+  fillIn('.project-name', 'Todos');
+
+  andThen(() => {
+    assert.deepEqual(redux.getState().project, { name: 'Chores' }, 'state should not be affected by Ember binding');
+  });
+});
--- a/tests/dummy/app/components/project-detail/component.js
+++ b/tests/dummy/app/components/project-detail/component.js
@@ -1,0 +1,19 @@
+import Ember from 'ember';
+import hbs from 'htmlbars-inline-precompile';
+import connect from 'ember-redux/components/connect';
+
+const { Component } = Ember;
+
+const stateToComputed = state => {
+  return {
+    project: state.project
+  };
+};
+
+const ProjectDetailComponent = Component.extend({
+  layout: hbs`
+  {{project-edit project=project}}
+  `
+});
+
+export default connect(stateToComputed)(ProjectDetailComponent);
--- a/tests/dummy/app/components/project-edit/component.js
+++ b/tests/dummy/app/components/project-edit/component.js
@@ -1,0 +1,10 @@
+import Ember from 'ember';
+import hbs from 'htmlbars-inline-precompile';
+
+const { Component } = Ember;
+
+export default Component.extend({
+  layout: hbs`
+  <input class="project-name" value={{project.name}} oninput={{action (mut project.name) value="target.value"}}>
+  `
+});
--- a/tests/dummy/app/project/template.hbs
+++ b/tests/dummy/app/project/template.hbs
@@ -1,0 +1,1 @@
+{{project-detail}}
--- a/tests/dummy/app/reducers/index.js
+++ b/tests/dummy/app/reducers/index.js
@@ -6,6 +6,7 @@ import roles from 'dummy/reducers/roles';
 import items from 'dummy/reducers/items';
 import list from 'dummy/reducers/list';
 import models from 'dummy/reducers/models';
+import project from 'dummy/reducers/project';
 
 export default {
   low: low,
@@ -15,5 +16,6 @@ export default {
   roles: roles,
   models: models,
   items: items,
-  list: list
+  list: list,
+  project: project
 };
--- a/tests/dummy/app/reducers/project.js
+++ b/tests/dummy/app/reducers/project.js
@@ -1,0 +1,10 @@
+const initialState = {
+  name: 'Chores'
+};
+
+export default (state=initialState, action) => { // jshint ignore:line
+  if (action.type === 'UPDATE_PROJECT') {
+    return Object.assign(state, action.data);
+  }
+  return state;
+};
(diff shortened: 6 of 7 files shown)


toranb:
@dustinfarris I always use `Object.assign({}, boundState, newObject);` because this avoids mutation of boundState (rule number 1 when using redux). I just referenced the egghead.io series and it clearly states this is how reducers should work

https://egghead.io/lessons/javascript-redux-avoiding-object-mutations-with-object-assign-and-spread

If people run into an error (as you show with this failing test), that will be a good thing because people shouldn't be using `Object.assign(boundState, newObject);` as it would alter the boundState making it impossible to time travel /rollback to a previous state of the application.

toranb:
@dustinfarris another resource I wanted to mention here is that I use deepFreeze when unit testing my reducer to catch these errors quickly. The idea is that if the incoming state is modified in any way the test will blow up like so

`TypeError: Cannot assign to read only property '??' of object...`

https://github.com/toranb/ember-redux-yelp/blob/master/tests/unit/reducers/results-test.js#L59

dustinfarris:
@toranb you're right, my example was bad form in general.  Thanks for the pointers on using freeze in unit tests.  I will do this on my own going forward.

I still think there is a possible "bug" here in that we allow Ember to bind to redux state (which mutates it, albeit subtly, by adding meta data, getters, setters, etc.).

I have updated my failing test (see latest commit) to show a better example of how this can bite you if you are not careful.

In the example here, the user sets up an input that mutates an attr it was given.  Perhaps the author of this "presentation" component is not aware that the attr came from redux.  If the user mutates the property, redux state is also mutated.

```hbs
<input value={{project.name}} oninput={{action (mut project.name) value="target.value"}}>
```

See the failing test that shows how this immediately affects redux state.

I'm sure there are many other examples where this or something similar can happen.  I guess the question is: upon whom does the burden fall to ensure that this sort of binding can/does not occur?

Either the user must take pains to use buffers and/or deepCopy everything, or we can potentially do this for the user when we define the properties.  Or maybe something else I haven't thought of.

brettburley:
> If the user mutates the property, redux state is also mutated. 

Does this not happen in redux as well? I would expect that each component in react-redux wouldn't make a deep copy of each prop it receives from the store.

That's not to say ember-redux shouldn't do better if it can. @toranb had mentioned this before in the context of testing against glimmer 2.0, but it would be great if we had a data-heavy test bed for testing performance impacts of changes like this. I don't have great intuition on how this may impact a large app with potentially many components being rendered and a large state tree.

Nonetheless, I do agree that the mutation from Ember for setting up the observers is less than ideal.

toranb:
@dustinfarris you are correct that ember modifies the POJO and I'd obviously prefer ember not do that in this scenario because we always proxy the POJO w/ a true ember computed in the component (meaning the POJO isn't ever required to have getters like you see today).

The input example you show is a great way to see this fall down, but again it's not good practice to use the mut helper w/ any redux state as you shouldn't mutate it. In that regard, this error is actually really helpful to warn/prevent the user from making a mistake (as they should be doing DDAU instead and using Object.assign in the reducer)

The negative here is we see the POJO has bindings attached but in my experience those are mostly simple meta/getters that don't interfere with time travel or redux in general. If we did remove these we have pros/cons to consider. 

Pro: removing it *might make the application more performant as you won't need to apply and watch any attrs on the POJO. To honestly confirm this we still need a performance example app of some kind to show it off.

Con: people using the mut helper in templates/ or those who try to modify state directly will not receive an error message unless we augment it in some way.

Pro/Con: Doing a deep copy to avoid this then requires we do some logic/work to prevent users from mutating state and warn them. Today we do nothing and get this behavior with what looks like no true downside to speak of (other than if you inspect the object in the reducer you see getter/setters that are never used)

dustinfarris:
@toranb to clarify, there is no error thrown on the input example.  Redux state is simply mutated.

I think at the very least we should be up front about this nuance in our documentation so that users are educated on how to program defensively.  I was personally not previously aware that Ember binding worked this way, and it took me a while to figure out what was going on when things started behaving unexpectedly.

toranb:
@dustinfarris right on. Today I'm honestly in a hold pattern regarding this issue myself. I see adding bindings isn't good at all but I'm also not excited to add yet another allocation (by doing a deep copy or something equivalent).

If we don't change anything on our side, what documentation or clarification should be made to protect developers as you mention? In the input example above I just assume people know not to (mut) anything as you shouldn't ever alter state (and only transform in the reducer). Because of this heavy bias I'm uncertain how to word this/ what you are asking for explicitly.

dustinfarris:
I'll write something up in a PR.  Basically to remind users about Ember 2-way binding and how it can affect redux state

toranb:
@dustinfarris sounds good - for now I'm planning to close this until a future date in which we decide to stand firmly on one side or the other (after watching glimmer components land/ etc)

PR #59, opened 2016-12-07
Coalesce changes in handleChange

scottmessinger:
By wrapping the changes in `beginPropertyChanges` and `endPropertyChanges`, we should be able to achieve the same performance wins as using `setProperties` (https://github.com/emberjs/ember.js/blob/v2.10.0/packages/ember-metal/lib/set_properties.js#L5-L7). A caveat: {begin,end}PropertyChanges are private methods.

### DIFF
--- a/addon/components/connect.js
+++ b/addon/components/connect.js
@@ -5,7 +5,9 @@ const {
   defineProperty,
   inject: { service },
   isEmpty,
-  run
+  run,
+  beginPropertyChanges,
+  endPropertyChanges
 } = Ember;
 
 export default (stateToComputed=() => ({}), dispatchToActions=() => ({})) => {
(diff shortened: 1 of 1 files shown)


toranb:
@scottmessinger totally cool with additional perf improvements but before we merge something like this I need to see how it performs before/and after (to know it's valuable/ not a premature or non effective optimization).

Would you be up for creating a perf /benchmark app with ember-redux so we can take these type of PRs into consideration? I personally have my eye on the complex (many components) demo from EmberConf 2016 but hadn't found the bandwidth to pull up the source and hack out the ember-redux version

https://github.com/tildeio/glimmer/blob/master/demos/uptime-boxes.html

toranb:
@scottmessinger worth a mention - the up coming react-redux v5 release includes a connectAdvanced function that can be used for components that don't need a callback/re-render fired on each change of the redux state. I plan to look at the final implementation of this when it ships late this month/early next and see what it would take to land that in ember-redux

https://github.com/reactjs/react-redux/blob/next/src/components/connectAdvanced.js#L43

toranb:
@scottmessinger I'm closing this out for now but I'm 100% open to a perf rework if and when we have something more mature in place to prove this is helpful/valuable

PR #60, opened 2016-12-08
[FEATURE]: the component instance is now available as `this` for non [phat]Arrow

toranb:
@foxnewsnetwork was asking for a 3rd argument for `stateToComputed` that would allow the user to pull from non attrs.

This example shows how I can pull a value from an injected service w/ the api change mentioned.

The next PR that follows this should likely add attrs/ and the instance arguments to `dispatchToActions` (as react-redux supports `ownProps` and it would be consistent with `stateToComputed`)

### DIFF
--- a/addon/components/connect.js
+++ b/addon/components/connect.js
@@ -23,11 +23,11 @@ export default (stateToComputed, dispatchToActions=() => ({})) => {
       init() {
         const redux = this.get('redux');
 
-        let props = stateToComputed(redux.getState(), this.getAttrs());
+        let props = stateToComputed.call(this, redux.getState(), this.getAttrs());
 
         Object.keys(props).forEach(name => {
           defineProperty(this, name, computed(() =>
-            stateToComputed(redux.getState(), this.getAttrs())[name]
+            stateToComputed.call(this, redux.getState(), this.getAttrs())[name]
           ).property().readOnly());
         });
 
@@ -38,7 +38,7 @@ export default (stateToComputed, dispatchToActions=() => ({})) => {
         }
 
         this.actions = Object.assign({},
-          this.actions, dispatchToActions(redux.dispatch.bind(redux))
+          this.actions, dispatchToActions.call(this, redux.dispatch.bind(redux))
         );
 
         this._super(...arguments);
@@ -47,7 +47,7 @@ export default (stateToComputed, dispatchToActions=() => ({})) => {
       handleChange() {
         const redux = this.get('redux');
 
-        const props = stateToComputed(redux.getState(), this.getAttrs());
+        const props = stateToComputed.call(this, redux.getState(), this.getAttrs());
 
         const notifyProperties = Object.keys(props).filter(name => {
           return this.get(name) !== props[name];
--- a/tests/acceptance/rerender-test.js
+++ b/tests/acceptance/rerender-test.js
@@ -13,6 +13,20 @@ module('Acceptance | rerender test', {
   }
 });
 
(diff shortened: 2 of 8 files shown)


toranb:
Just found withRef option and I'll rework this to add that instead

foxnewsnetwork:
thanks mate, look forward to the `withRef` PR

foxnewsnetwork:
LGTM; while I absolutely agree `this` is an anti-pattern, it is necessary for the poor souls who are stuck on ember 1.12 and below who actually need redux the most to help clean up our code-base's state problems.

toranb:
cc danielchappell

toranb:
@foxnewsnetwork I've rebased this and it's ready for a look when you have time. Two things to note about this

1) if you are using a local (on the component) CP and need to cache bust on the redux prop (mapped when stateToComputed runs) you will be required to `notifyPropertyChange('<<mappedProp>>')` if you need it to re-evaluate /re-render

2) this isn't anything we will publicly mention, it will be a `intimate api` as I've often heard them called

Any other feedback welcome -thanks for discussion around this!

addon/components/connect.js
```diff
@@ -38,7 +38,7 @@ export default (stateToComputed, dispatchToActions=() => ({})) => {
         }
 
         this.actions = Object.assign({},
-          this.actions, dispatchToActions(redux.dispatch.bind(redux))
+          this.actions, dispatchToActions.call(this, redux.dispatch.bind(redux))
```
foxnewsnetwork: 👍 this is particularly useful for the sad souls stuck in ember <1.11 land; `(action something)` isn't a helper before the 1.13 so very often we have to depend on having `this` to get variables out. Hooray

toranb: @foxnewsnetwork can you clarify the part you mention ^^ `variables out` ? I have a failing local build trying to use a computed property (as it would require a "break cache" like mechanic and I can't get that to work as-is). I just wanted to learn more about 1 or 2 use cases you might have to ensure they will actually "work" once we merge it.

toranb:
@foxnewsnetwork I pushed up a second commit to this PR showing how I'd use a local CP w/ the stateToComputed function but it only works because of the `didUpdateAttrs` hook (that in turn does the notifyPropertyChange). So ... if the thing you key off isn't an attr I'm not certain this will be of much value.

What use case did you have in mind? I'd like to document it here w/ a test even. I'm hesitant to merge something that won't work for you (sorry for the less than confident tone here -not using this myself makes me a little unsure of myself)

foxnewsnetwork:
The primary use case is when you have something like this:

`app/services/session.js`
```javascript
Ember.Service.extend({
  isLoggedIn: false,
  checkLoginStatus: Ember.observer('isLoggedIn', function() { 
    // terrible state-mutation code
  })
});
```

and in a component, I'd like to be access `isLoggedIn` via the injected service; say something like:

```javascript
function statesToCompute(state, attrs) { 
  if(this.get('isLoggedIn') {
    return { ... };
  } else { 
    return { ... };
  }
}
```

IMHO, though having components that require a ton of ember injections is bad code on my part, and ember-redux shouldn't have to bend-over backwards for my inability to not write awful code. My current solution is to pass in `isLoggedIn` to the component so it shows up in the attrs parameter. Using `call` and passing in `attrs` is already excellent, so this LGTM

toranb:
@foxnewsnetwork thanks for the much needed feedback on this PR! It seems the attr approach you are using will work so I'm planning to merge this unless I hear otherwise.

I plan to publish v1.11 later today and it should include this :)

PR #61, opened 2016-12-08
API docs

dustinfarris:
**WORK IN PROGRESS**

## Summary

This is a first stab at API documentation.

## Details

To get the ball rolling, I've filled out docstrings in connect, but want feedback on what our exact strategy is going to be in terms of hosting.  I looked at yuidoc first, but it is heavily biased to classes.  So I went with good-ol [JSDoc](http://usejsdoc.org/index.html) which is working well.

You can use a new npm command to build the docs:

```
npm docs
```

They will be available in the out/ directory which has been added to gitignore.

## Considerations

- Where do we host this?  How do we automate deployment?
- How does this work with the current ember-redux.com site?
- Due to the function-returning-a-function-returning-a-function implementation for connect, JSDoc does not pick up the inner docstrings for the connected component.  This is probably ok since that code should be considered private anyway.

## Alternative

We could instead add a new API.md file to the gh-pages branch.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 /bower_components
 
 # misc
+/out
 /.sass-cache
 /connect.lock
 /coverage/*
--- a/README.md
+++ b/README.md
@@ -31,11 +31,10 @@ import Ember from 'ember';
 import hbs from 'htmlbars-inline-precompile';
 import connect from 'ember-redux/components/connect';
 import ajax from 'example/utilities/ajax';
-import getUsersByAccountId from '../reducers';
 
 var stateToComputed = (state, attrs) => {
   return {
-    users: getUsersByAccountId(state, attrs.accountId)
+    users: state.users.all
   };
 };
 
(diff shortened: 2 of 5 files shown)


toranb:
@dustinfarris love the comments! One of the reasons I created `ember-redux.com` was to avoid hosting a website/paying for hosting. Could we just run this npm command with each release and update the gh-pages side with the output? Or did you want something more complex?

Once we workout these details we can discuss what it looks like to script the deploy.

dustinfarris:
@toranb yeah i think that might work!  will need to do some git trickery since i believe gh-pages is an orphaned branch—but that's what a lot of those blog engines do so shouldn't be too hard.

We might be able to set up Travis to do this for every successful build to master—what do you think?

dustinfarris:
Well, this is good timing: 

https://github.com/blog/2289-publishing-with-github-pages-now-as-easy-as-1-2-3

dustinfarris:
@toranb I think we can move everything to a /docs folder on master now with the new GitHub publishing.  I'm excited to see what this looks like.

Not sure how to integrate with JSDoc though.  We might want to ditch that idea after all if GitHub is going to make the alternative so easy.

dustinfarris:
Going to resubmit this in conjunction with moving docs from gh-pages to /docs on master

PR #67, opened 2016-12-14
Allow `null` stateToComputed argument

dustinfarris:
This restores the ability to pass `null` for `stateToComputed` when connecting a component.

### DIFF
--- a/tests/integration/connect-test.js
+++ b/tests/integration/connect-test.js
@@ -93,3 +93,25 @@ test('lifecycle hooks are still invoked', function(assert) {
 
   this.set('name', 'Dustin');
 });
+
+test('connecting dispatchToActions only', function(assert) {
+  assert.expect(2);
+  const dispatchToActions = () => {};
+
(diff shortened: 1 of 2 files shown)


tests/integration/connect-test.js
```diff
@@ -93,3 +93,17 @@ test('lifecycle hooks are still invoked', function(assert) {
...
+
+test('connecting dispatchToActions only', function(assert) {
+  assert.expect(1);
+  const dispatchToActions = () => {};
+
+  this.register('component:test-component', connect(null, dispatchToActions)(Ember.Component.extend({
+    init() {
+      this._super(...arguments);
+      assert.ok(true, 'should be able to connect components passing `null` to stateToComputed');
+    }
```
toranb: @dustinfarris can we also add something here for undefined (assuming both work obviously - just want to keep us in check going forward)

dustinfarris: Added!

PR #69, opened 2016-12-14
Reduce the number of run loops created when state changes

### LINKED ISSUE #64: [Ember 2.10] Uncaught Error: infinite rendering invalidation detected
2.10 introduced Glimmer 2. Glimmer 2 introduced [infinite rendering invalidation detection](https://github.com/emberjs/ember.js/blob/855363285dc66ece8eac9f417375bd4d69460e7e/packages/ember-glimmer/lib/renderer.js#L149-L163). Not entirely sure what that is, haven't dug enough.

We have a master detail layout in our app, and the detail panel is what uses redux. We have it carved out as an addon.

When a user clicks the first item, none of the detail components exist yet. The detail panel slides in, the detail addon top level component is rendered with a handful of inputs, it makes a few server calls and renders all the things.

When a user clicks a second item, because the panel is already open, we have ~15ish components already bound to state and listening for updates. Clicking the second item passes in a bunch of inputs which are immediately `dispatch({ type: INITIALIZE: payload: inputs})` into state.

There are enough components listening to some of the inputs that `handleChange` is called a good number of times.  Too many.  And then that error happens and things explode.

There are definitely steps I can take to reduce the number of components bound into state.  I can also do some work to [batch some actions](https://github.com/tshelburne/redux-batched-actions) so that synchronous actions that happen to have some overlapping state do not get updated multiple times.

But I can absolutely foresee, post improvements, still having this issue and being forced to destroy most all of the detail before I do anything with the new inputs so that components aren't listening for changes. In some cases this might be the ideal thing to do anyway, and I expect that'll be one of the best improvements I can make, but in some it isn't ideal.

Not sure exactly what ember-redux can do differently to smooth this over, but I'd guess it'll be something other folks run into at some point.

dbashford:
This fixes #64 

Starting with `2.10`, we've begun having issues with excessive and spammy `run` usage (and not just with ember-redux).

The `run(() => this.handleChange())` in particular was causing us problems.  In a relatively small app with 7-10 components getting 4 or 5 quick state updates in succession, we were hitting that `run` 50+ times in a short period of time. With `2.10` that consistently gave us an [`infinite rendering invalidation detected`](https://github.com/emberjs/ember.js/blob/855363285dc66ece8eac9f417375bd4d69460e7e/packages/ember-glimmer/lib/renderer.js#L156) error.

There isn't a lot of information about that particular error, and the Ember code is a bit obtuse, but based on a lot of experimentation the past few days it seems closely related to excessive run loop creation. This PR aims to curb that.

The following is a `console.log` of the`notifyProperties` for a group of components after a button click kicked off a series of redux actions. Every time there is an empty array, that's one `run` avoided.

image

Even if this weren't a solution to a specific problem we are having, it seems a decent perf improvement. No need to involve Ember run loops if there's nothing to do in the end?

I tried altering this PR code to use [`run.join`](http://emberjs.com/api/classes/Ember.run.html#method_join) rather than `run` as it feels like that would further aid in not creating new run loops. It was successful in my app. But a rerender test broke requiring [a dispatch](https://github.com/toranb/ember-redux/blob/e924d0a6adafeed339272aee9857319c3a01ff62/tests/acceptance/rerender-test.js#L25)  be wrapped in a `run` to pass. Decided to leave that out.

### DIFF
--- a/addon/components/connect.js
+++ b/addon/components/connect.js
@@ -33,7 +33,7 @@ export default (stateToComputed, dispatchToActions=() => ({})) => {
 
         if (!isEmpty(Object.keys(props))) {
           this.unsubscribe = redux.subscribe(() => {
-            run(() => this.handleChange());
+            this.handleChange();
           });
         }
 
@@ -47,13 +47,17 @@ export default (stateToComputed, dispatchToActions=() => ({})) => {
       handleChange() {
         const redux = this.get('redux');
 
-        let props = stateToComputed(redux.getState(), this.getAttrs());
+        const props = stateToComputed(redux.getState(), this.getAttrs());
 
-        Object.keys(props).forEach(name => {
-          if (this.get(name) !== props[name]) {
-            this.notifyPropertyChange(name);
-          }
+        const notifyProperties = Object.keys(props).filter(name => {
+          return this.get(name) !== props[name];
         });
+
+        if (notifyProperties.length > 0) {
+          run.join(() => {
+            notifyProperties.forEach(name => this.notifyPropertyChange(name));
+          });
+        }
       },
 
       /**
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,7 @@
 /libpeerconnection.log
 npm-debug.log*
 testem.log
+
+# vscode
+jsconfig.json
+typings
--- a/tests/integration/runloop-test.js
+++ b/tests/integration/runloop-test.js
@@ -34,17 +34,17 @@ test('handleChange is invoked inside runloop explicitly', function(assert) {
   assert.equal($parent.text(), 1);
 });
 
-// test('handleChange will join an existing runloop when exists', function(assert) {
-//   this.render(hbs`{{count-list}}`);
+test('handleChange will join an existing runloop when exists', function(assert) {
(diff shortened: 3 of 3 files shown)


addon/components/connect.js
```diff
@@ -43,13 +43,17 @@ export default (stateToComputed=() => ({}), dispatchToActions=() => ({})) => {
...
-        Object.keys(props).forEach(name => {
-          if (this.get(name) !== props[name]) {
-            this.notifyPropertyChange(name);
-          }
+        const notifyProperties = Object.keys(props).filter(name => {
+          return this.get(name) !== props[name];
         });
+
+        if (notifyProperties.length > 0) {
+          run(() => {
```
runspired: this should very likely be `run.join`

toranb:
@runspired can you articulate when run.join should be used and why? I do agree with your comment - just wanted to be educated if at all possible

toranb:
@dbashford I merged the PR that better supports run.join for tests. You can rebase and uncomment the last test in runloop-test.js to drive out the last set of changes for this pull request

dbashford:
Rebased, switched to `join` and uncommented the test. Thanks for knocking that out for me!

runspired:
Yay! :)

PR #70, opened 2016-12-15
[TESTS]: improved runloop testing to confirm both run/and run.join work

toranb:
@dbashford if we can merge this quick it will provide a simple way for you to verify the run.join works on your branch. I tested this w/ your fork and it was g2g

If someone lands me a LGTM+1 I'll merge and @dbashford can rebase /uncomment the test

### DIFF
--- a/tests/integration/runloop-test.js
+++ b/tests/integration/runloop-test.js
@@ -1,0 +1,50 @@
+import Ember from 'ember';
+import hbs from 'htmlbars-inline-precompile';
+import { moduleForComponent, test } from 'ember-qunit';
+
+var original, joined;
+
+moduleForComponent('count-list', 'integration: runloop test', {
+  integration: true,
+  setup() {
+    this.inject.service('redux');
+
+    joined = false;
+    original = Ember.run.join;
+    Ember.run.join = function() {
+      joined = true;
+      return original.apply(this, arguments);
+    };
+  },
+  teardown() {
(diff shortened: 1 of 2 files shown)


toranb:
@dustinfarris any feedback on this change? it still fails w/out run (your main concern a month or so back). It also has a test that works fine with both run and run.join (actually fails w/out run.join -the reason it's commented out).

dustinfarris:
@toranb my concerns a month ago stem from my taking the documentation surrounding Ember.run.* at face value, and I would imagine that wrapping an entire third-party async callback in Ember.run is probably OK the majority of the time—more so on the app level.  But here we are in an addon that is seeing some incredible usage that pushes the limits.  While probably not a priority, I think additional use-case examples would be helpful in the Ember.run docs.

tl;dr, given new information and input from an Ember core team member, I'm all for embracing run.join here.

tests/integration/runloop-test.js
```diff
@@ -0,0 +1,50 @@
...
+  let $parent = this.$('.parent-state');
+
+  assert.equal($parent.text(), 0);
+
+  this.redux.dispatch({type: 'UP'});
+
+  assert.equal($parent.text(), 1);
+});
+
+// test('handleChange will join an existing runloop when exists', function(assert) {
```
dustinfarris: is this a place-holder?

PR #75, opened 2016-12-29
[BUILD]: updated travis and code climate badges

toranb:
Previously these badges pointed at the toranb url


toranb:
@dustinfarris merging this in w/out the new process to get the build badges and such updated asap

dustinfarris:
oh i don't think any of that is formalized yet.  i figure we'll reopen that when we get to the docs.

you're still BDFL as far as i'm concerned 🙌

PR #77, opened 2017-01-02
Yarn

dustinfarris:
Just adding the lockfile.  This doesn't affect end-users.  If you are a contributor and you're using yarn, you can now:

```
yarn install && bower install
```

instead of using npm, if you so desire.


toranb:
I always heard yarn was for apps and npm was for libraries/addons. Thoughts from other people on the subject?

dbashford:
I've not heard that, but I can sympathize with it. I've become a fan of Yarn for its speed since I introed it to our app builds. Sped us up quite a bit and gave us a cache we could rely upon. It's a PITA that `ember install` uses `npm install` under the hood which means the lock file is bypassed. We've had to add a precommit hook that catches when someone is updating `package.json` without a `yarn.lock` update.

Considering the builds here aren't a problem and we aren't re-installing node modules over and over, the benefits are minimal IMHO.

That said, including the `yarn.lock` file doesn't mean anyone has to use it. Just helps those that do.

toranb:
@dbashford https://github.com/ember-cli/ember-cli/issues/6341

About the application vs library - read the section titled "Libraries vs Applications"

https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all

MiguelMadero:
@toranb I think it still makes sense for libs. See the section where James mentions the contributor's burden.

toranb:
@MiguelMadero fair enough - I just don't see or hear of other addon authors using yarn. Are you using it internally for private addons you maintain?

dustinfarris:
It's slowly but surely getting added to everything as far as I can tell.

E.g. react-redux  https://github.com/reactjs/react-redux/blob/master/yarn.lock

MiguelMadero:
Yes. For everything internally and plan to move my addons on the next
opportunity. Also plan to do it for travis, but run yarn update to catch
breaking changes.

toranb:
@dustinfarris sounds like this is somethjng we can merge this weekend. I'm planning the v1.11 release and would like to include this if possible

dustinfarris:
Do you want to add to Travis?

toranb:
@dustinfarris should we wait until ember-cli supports it by default or no ?

https://github.com/ember-cli/ember-cli/issues/6341

dustinfarris:
@toranb hmm, i don't think it matters, and i'm already adding to CI in other addons.  I like @MiguelMadero's idea to run `yarn update` every time to catch breaks in new dependency versions.

There is logic in keeping the CI install close to the (officially supported) ember-cli install.

I don't think there's a right answer.  My preference is to go all-in with yarn.

toranb:
@dustinfarris left this out of the v1.11 release for now. Wanted more time to see what a (more) mature travis ci stack looks like with yarn :)

toranb:
@MiguelMadero @dbashford quick update - looks like ember-cli v2.13X will start pushing yarn so I'd assume addons will "play nice" with yarn by default once we upgrade to that stable release. Anyone heard otherwise?

If so, I'll reopen this issue (and link it) once I've upgraded the addon itself to ember-cli v2.13+ (in the coming weeks). After that we could add a yarn.lock and roll right into the future :)

PR #98, opened 2017-02-18
Support non-function properties in route

sandersky:


sandersky:
@dustinfarris I'm actually not using controllers at all (or rather not creating `controller.js`, just using what Ember auto-generates at runtime). Instead my route dispatches an action to fetch the initial list data in the `model()` hook, and all subsequent updates for paging, searching, sorting, etc occur within the component, which dispatches actions to the redux store. The route is then observing the store for changes and updating it's queryParams when appropriate. This makes for a better separation of concerns as the component could care less about query parameters and whether or not it is being rendered in it's own route. The only piece currently strongly tying the component to the route is the fact that the route makes the initial fetch.

```js
import route from 'ember-redux/route'

import {updateListItems} from 'my-app/actions/items'

function getQueryParamsFromReduxStoreState (state) {
  return {
    limit: state.limit || 10,
    page: state.page || 0,
    query: state.query || '',
    sort: state.sort || 'name'
  }
}

export default route({
  queryParams: {
    limit: {refreshModel: false},
    page: {refreshModel: false},
    query: {refreshModel: false},
    sort: {refreshModel: false}
  },

  activate (dispatch, ...args) {
    this._super(...args)
    this.subscribeToReduxStore()
  },

  deactivate (dispatch, ...args) {
    this._super(...args)
    this.unsubScribeFromReduxStore()
  },

  model (dispatch, params, {queryParams}) {
    const {limit, page, query, sort} = queryParams

    dispatch(
      updateListItems({limit, page, query, sort})
    )
  },

  onReduxStoreChange (dispatch) {
    const state = this.get('redux').getState().listView
    const currentQueryParams = this.get('router.router.state.queryParams')
    const expectedQueryParams = getQueryParamsFromReduxStoreState(state)
    const expectedKeys = Object.keys(expectedQueryParams)

    if (expectedKeys.some((key) => currentQueryParams[key] !== expectedQueryParams[key])) {
      this.transitionTo(this.routeName, {
        queryParams: expectedQueryParams
      })
    }
  },

  subscribeToReduxStore () {
    this._unsubscribe = this.get('redux')
      .subscribe(this.onReduxStoreChange.bind(this))
  },

  unsubScribeFromReduxStore () {
    if (!this._subscription) return
    this._unsubscribe()
    this._unsubscribe = null
  }
})()
```

addon/route.js
```diff
@@ -1,17 +1,21 @@
...
     return WrappedRoute.extend({
       redux: Ember.inject.service('redux'),
       init() {
         var redux = this.get('redux');
         var route = this;
-        Object.keys(funcs).forEach(function(func) {
-          route[func] = function(...args) {
+        Object.keys(props).forEach(function(key) {
+          if (typeof props[key] !== 'function') {
+            route[key] = props[key]
```
toranb: semicolon on lines 13/14? :)

toranb: also ... why didn't eslint complain during the build? should we alter that as part of this PR to ensure less bike shedding in future PRs?

tests/dummy/app/query-params/route.js
```diff
@@ -0,0 +1,21 @@
+import Ember from 'ember';
+import route from 'ember-redux/route';
+
+var queryParams = {
+  foo: {
+    refreshModel: false
+  }
```
toranb: why set refreshModel to false here? trying to read thru everything to better understand this so forgive me if I've totally missed it

tests/dummy/app/query-params/route.js
```diff
@@ -0,0 +1,21 @@
...
+function afterModel(dispatch, model, transition) {
+  if (Ember.isPresent(transition)) {
+    dispatch({
+      queryParams: Object.keys(this.get('queryParams')),
+      type: 'SET_QUERY_PARAMS'
+    });
+  }
+}
+
+var QueryParamsRoute = Ember.Route.extend();
```
toranb: you shouldn't need this line - if you only write functions like this our `route` helper will provide a full `Ember.Route` for you (like ember generates controllers when you don't need to override anything)

toranb: coming back to this now after reading the entire PR it seems we should maybe start on another route first (in the acceptance test) to assert the URL (assuming we can ensure no query params are present doing this?). Next we would transition to the query-params route (in which case the after model would fire) -asserting the diff after the dispatch fires. Finally what I feel is missing (maybe?) is that we should alter the query param in some way and see if that is correctly updated/reflected e2e. Could you maybe add a button to the child component and click it altering the queryParam in some way so we can see it's updated (executing the flow thru the reducer with a new value) ?

Maybe I just missed all that but from the test it seems we only flow thru 1x and I'd love to see us update from the default value if possible.

Ping me on slack today if I'm not clear :)

Also - thank you 100x for this pull request! Taking real world pain and patching it will save others pain down the road! thank you, thank you, thank you!

brettburley:
I probably missed some prior discussion on this, but is this functionally equivalent to the syntax you'd use today? E.g.
```
import Ember from 'ember';
import route from 'ember-redux/route';

function afterModel(dispatch, model, transition) {
  if (Ember.isPresent(transition)) {
    dispatch({
      queryParams: Object.keys(this.get('queryParams')),
      type: 'SET_QUERY_PARAMS'
    });
  }
}

export default route({ afterModel })(Ember.Route.extend({
  queryParams: {
    foo: {
      refreshModel: false
    }
  }
}));
```

toranb:
I'm closing this out but I am reopening a PR w/ the tests + queryparams connect scenario (all code from this original PR).

I spent a good hour looking over the work here and decided not to pull this into the project. When I wrote this project the route helper was mostly a minimalist function that returned a new route w/ whatever model hook function(s) you passed in. I didn't want to hold state or do much of anything else for fear of having to maintain a fork in the road.

In situations like this (what Matt showed in the original PR) I would just build a normal ember route/ inject the redux service and use dispatch. This is a little more code but also it requires no work to keep it current. The future looks bright for service injection and this seems to be the happy path :)

That said, I did like the example Matt threw together and decided to PR 90% of his great work :)

PR #99, opened 2017-03-01
Support redux-devtools in browser

sandersky:
Support this: https://github.com/zalmoxisus/redux-devtools-extension


toranb:
I prefer this as an opt in like we mention in the readme. Why not go about it in that way?

sandersky:
@toranb I must have missed that in the README but I'm curious why you prefer opt-in rather than it just working out of box. If you don't have the devtools installed it'll just do what it always did before given the code `window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose`. I know in my case I'm going to be trying to convince a team to start using `ember-redux` and the more setup they have to do, the less likely they'll be wanting to make the switch. I'd say my argument for adding it directly into this project is for better DX (developer experience).

toranb:
@sandersky my only counter point is that the service is already (overly) complex and having the ability to "plugin" a middleware simply by adding a file feels better from a SRP /wrap my head around the code perspective. Another option (instead of what you have here) would be to ship a default middleware (with what you see in the README today) so upon `ember install` it's just up and running (without adding any lines of code or additional complexity to the service itself)

toranb:
s/middleware/enhancer/

my fault - used the wrong word :)

sandersky:
@toranb shall I add a blueprint that ships a middleware with the devtools setup?

jkarsrud:
If I can chime in, I like the idea of having this supported by default. A lot of people I know that use redux have this as part of their boilerplate setup, and always use it, so it's a nice thing to enable by default. What if it implements this by default through the blueprints instead of a separate addon?

sandersky:
@toranb I think it is worth keeping in the README in case someone already consuming the addon just bumps the version in their `package.json` without re-running blueprints on upgrading.

toranb:
@sandersky if we push what you have into `addon/enhancers` and export it from `app/enchancers` users would get for free w/out a blue print run (allowing us to remote that from the README). Also I think this is the preferred approach generally (for addon authors). We don't do that for the redux service here because of a custom hook we use to set everything up - did that block this/ break it? If not could we move this to the addon and export it as you often see?

If I'm wrong -totally up to hear about it :)

toranb:
@sandersky I think we can cut the readme and merge this. You mind updating the PR to alter the README in the next few days?

sandersky:
@otanb just updated the README to say install the browser plugin and not the rest.

README.md
```diff
@@ -86,16 +86,7 @@ export default UserTableComponent;
...
-
-3. In that file, add the following 3 lines of code:
-
-    ```js
-    //app/enhancers/index.js
-    import { compose } from 'redux';
-    var devtools = window.__REDUX_DEVTOOLS_EXTENSION__ ? window.__REDUX_DEVTOOLS_EXTENSION__() : f => f;
-    export default compose(devtools);
-    ```
+2. Enojy!
```
toranb: typo :)

toranb:
@sandersky rebase this when you have time - I just merged the ember 2.11.3 fix

sandersky:
@toranb fixed the typo and pulled in the upstream changes.

PR #100, opened 2017-03-04
DRY props

lolmaus:


lolmaus:
Ah, I've realized it gets an outdated state this way. Sorry.

PR #101, opened 2017-03-08
Bind action creators when dispatchToActions is an object

jkarsrud:
This adds another thing that is possible to do in `react-redux`; pass an object to `dispatchToActions` and have them automatically bound to dispatch. This allows you to create action creators as separate functions, and have them automatically be dispatched instead of wrapping them in a `dispatch` call yourself.

Below is an example component after this change:

```js 
import { createPost } from 'module/action-creators';

const mapDispatchToActions = {
  createPost
};

// compared to 
/*
const mapDispatchToActions = (dispatch) => {
  createPost: () => dispatch(createPost())
});
*/

export default connect(undefined, mapDispatchToActions)(Ember.Component.extend({
  layout: hbs`
  <button onclick={{action "createPost"}}>Create post</button>
  `
}));
```

I've found myself using this exclusively in React, and I personally think it's nice to get rid of some boilerplate — of which there is already quite a bit when using Redux.

### DIFF
--- a/addon/components/connect.js
+++ b/addon/components/connect.js
@@ -1,4 +1,5 @@
 import Ember from 'ember';
+import { bindActionCreators } from 'redux';
 
 const {
   computed,
@@ -37,9 +38,15 @@ export default (stateToComputed, dispatchToActions=() => ({})) => {
           });
         }
 
-        this.actions = Object.assign({},
-          this.actions, dispatchToActions.call(this, redux.dispatch.bind(redux))
-        );
+        if (typeof dispatchToActions === 'function') {
+          this.actions = Object.assign({},
+            this.actions, dispatchToActions.call(this, redux.dispatch.bind(redux))
+          );
+        }
+
+        if (typeof dispatchToActions === 'object') {
+          this.actions = Object.assign({}, this.actions, bindActionCreators.call(this, dispatchToActions, redux.dispatch.bind(redux)));
+        }
 
         this._super(...arguments);
       },
--- a/tests/integration/connect-test.js
+++ b/tests/integration/connect-test.js
@@ -136,3 +136,38 @@ test('connecting dispatchToActions only', function(assert) {
   this.render(hbs`{{test-component-1}}`);
   this.render(hbs`{{test-component-2}}`);
 });
+
+test('connecting dispatchToActions as object should dispatch action', function(assert) {
+  assert.expect(2);
+
+  const dispatchToActions = {
+    up() {
+      assert.ok(true, 'should be able to pass object of functions to dispatchToActions');
+      return {
+        type: 'UP'
+      };
+    },
+    down() {
+      assert.ok(true, 'should be able to pass object of functions to dispatchToActions');
+      return {
+        type: 'DOWN'
+      };
+    }
+  };
+
+  this.register('component:test-dispatch-action-object', connect(undefined, dispatchToActions)(Ember.Component.extend({
+    init() {
+      this._super(...arguments);
+    },
+    layout: hbs`
+      <button class="btn-up" onclick={{action "up"}}>up</button>
+      <button class="btn-down" onclick={{action "down"}}>up</button>
+    `
+  })));
+
+  this.render(hbs`{{test-dispatch-action-object}}`);
+  Ember.run(() => {
+    this.$('.btn-up').trigger('click');
+    this.$('.btn-down').trigger('click');
+  });
+});


toranb:
@dbashford could you carve out a little time to look this over in the next 5/7 days? Would love someone more familiar than me looking at this. I know it has been missing from the connect api since the very start so I want to be sure we solve it as you would expect

toranb:
@jkarsrud great work! This truly looks dead simple :)

toranb:
@jkarsrud the only request I have is to see you qire up one full flow acceptance test that shows the dispatch part in action (hitting a reducer to update some value / proving it will re render)

You mind adding a simple component like the up/down that just alters the number reducer we already use in other end-to-end test cases?

dbashford:
I'll take a look. May not be until next week.

dbashford:
And by next week I meant the week after next week. Finally digging into this now. Will let you know. =)

dbashford:
👍 

Looks good!

dbashford:
One little snag which bears mentioning and might trip someone up, but not sure it is a ember-redux problem.

Here's a cut down version of one of our `dispatchToActions`.

```javascript
const dispatchToActions = (dispatch) => ({
  toggleRequest: () => dispatch(VisualActions.toggleRequest()),
});
```

`toggleRequest` being activated via a click in the template: `click=(action "toggleRequest")`

Theoretically this turns into something like this:

```javascript
const dispatchToActions = { toggleRequest };
```

However, the `toggleRequest` accepts a parameter that allows for explicitly setting the request flag. In some cases we may have something like...

```javascript
const dispatchToActions = (dispatch) => ({
  requestOn: () => dispatch(VisualActions.toggleRequest(true)),
});
```

The problem with `const dispatchToActions = { toggleRequest };` is that `toggleRequest` will now get passed the `jQuery.Event` from the click.

`jQuery.Event`s probably ought not be handed to an action creator. If a dev on our team PRed something to the codebase where an `e` was being inspected in an action creator, the PR would get lit up. If the `jQuery.Event` must be inspected, I'd expect that to be handled in the component and for the result of that, some POJO or primitive, to be `this.send` into the action creator.

Not sure what we could/should even do in this case?

jkarsrud:
That is a snag for sure, but the problem is the same for React apps using react-redux, where you would need to wrap the action creator in a function that handles the event and calls the action creator.

So in Ember, you could make a new action that specifically handles the click event, and calls the bound action:

```js
handleToggleClick(e) {
  this.get('toggleRequest')(true);
}
```

It's perhaps something that's worth mentioning in docs when showing this shorthand?

On another note, I have some spare cycles for doing OSS this week, so hopefully I'll be able to write a component and proper tests for this to prove it actually works :)

dbashford:
Yep, alluded to the workaround with the action in the middle. And yeah, I'd recommend a mention in the docs. Lots of people will come here from Ember proper not having React+Redux experience, so it is worth being as clear with them as possible. "In this case a `jQuery.Event` is being sent into your Action Creator and you may not intend for that to be the outcome."

Given in most cases (like above) `bindActionCreators` is allowing me to replace a long one-liner with a short one-liner, when I have to write something like the below to use the short(er)hand, I'm just going to pass and provide the function rather than the object.

```javascript
handleToggleClick(e) {
  this.get('toggleRequest')(true);
}
```

toranb:
@jkarsrud appreciate any spare cycles here :)

@dbashford thanks for the feedback! totally legit to learn more about this (as I've not used action creators myself with ember/or react)

toranb:
@jkarsrud added the full e2e acceptance test over in a new PR

https://github.com/ember-redux/ember-redux/pull/107

Keeping you both in the loop here - I wanted to think about this jQuery event issue as I ran into it myself and I'd like a little more time to hack around on react-redux to see that in action.

addon/components/connect.js
```diff
@@ -37,9 +38,15 @@ export default (stateToComputed, dispatchToActions=() => ({})) => {
...
-          this.actions, dispatchToActions.call(this, redux.dispatch.bind(redux))
-        );
+        if (typeof dispatchToActions === 'function') {
+          this.actions = Object.assign({},
+            this.actions, dispatchToActions.call(this, redux.dispatch.bind(redux))
+          );
+        }
+
+        if (typeof dispatchToActions === 'object') {
+          this.actions = Object.assign({}, this.actions, bindActionCreators.call(this, dispatchToActions, redux.dispatch.bind(redux)));
```
brettburley: Looking at the `bindActionCreators` implementation, it seems like it use `this`, so it seems like it might be unnecessary to call it with the current context, like:
```
bindActionCreators(dispatchToActions, redux.dispatch.bind(redux))
```

brettburley: I think this also means that actions bound in this way don't get the `this` context, so they can't do things like `this.get` in their functions, which was discussed briefly in https://github.com/ember-redux/ember-redux/pull/60.

jkarsrud:
Closing this in favour of #107.

PR #107, opened 2017-04-13
[FEATURE]: unlock action creators

toranb:
I took the work @jkarsrud started and extended it to show a few points raised in the original [issue](https://github.com/ember-redux/ember-redux/pull/101). All credit for the action creator implementation goes to him - I just wanted to speed up some of this as it's possible I may use it in the near future.

1) I added acceptance tests that exercise it from end to end
2) I show the jQuery event issue mentioned by wrapping the action creator (still not sure how I feel about this ...)
3) I learned how this feature was intended to work by reading the [docs](http://redux.js.org/docs/recipes/ReducingBoilerplate.html) and used the "async action creator" as that would be how I'd likely use it in the wild

One issue I'm fighting with right now is that since the last middleware to boot is `redux-saga` (in the test suite) I get this fun error when I acceptance test it. Note: if I just run `ember s` it does indeed work as designed :) so I just need to hack a bit more until this darn middleware is setup correctly :)

screen shot 2017-04-12 at 9 12 54 pm

### DIFF
--- a/addon/components/connect.js
+++ b/addon/components/connect.js
@@ -1,4 +1,5 @@
 import Ember from 'ember';
+import { bindActionCreators } from 'redux';
 
 const {
   computed,
@@ -37,9 +38,16 @@ export default (stateToComputed, dispatchToActions=() => ({})) => {
           });
         }
 
-        this.actions = Object.assign({},
-          this.actions, dispatchToActions.call(this, redux.dispatch.bind(redux))
-        );
+        if (typeof dispatchToActions === 'function') {
+          this.actions = Object.assign({},
+            this.actions, dispatchToActions.call(this, redux.dispatch.bind(redux))
+          );
+        }
+
+        if (typeof dispatchToActions === 'object') {
+          this.actions = Object.assign({},
+            this.actions, bindActionCreators(dispatchToActions, redux.dispatch.bind(redux)));
+        }
 
         this._super(...arguments);
       },
--- a/tests/dummy/app/components/less-boilerplate/component.js
+++ b/tests/dummy/app/components/less-boilerplate/component.js
@@ -1,0 +1,23 @@
+import Ember from 'ember';
+import hbs from 'htmlbars-inline-precompile';
+import connect from 'ember-redux/components/connect';
+import { bumpTwice } from '../../actions/index';
+
+var stateToComputed = function(state) {
+  return {
+    low: state.low
+  };
+};
+
+var mapDispatchToActions = {
+  bumpTwice
+};
+
+var LessBoilerplateComponent = Ember.Component.extend({
+  layout: hbs`
+    <span class="upp-low">{{low}}</span>
+    <button class="btn-upp" onclick={{action "bumpTwice" 1}}>upp</button>
+  `
+});
+
+export default connect(stateToComputed, mapDispatchToActions)(LessBoilerplateComponent);
--- a/tests/acceptance/async-dispatch-test.js
+++ b/tests/acceptance/async-dispatch-test.js
@@ -166,3 +166,24 @@ test('connected routes provide an ember route for you by default', function(asse
     assert.equal(find('.user-name').length, 2);
   });
 });
+
+test('connect supports action creator syntax', function(assert) {
+  visit('/actionz');
+  andThen(() => {
+    assert.equal(currentURL(), '/actionz');
+    assert.equal(find('.upp-low').text(), '0');
+  });
+  click('.btn-upp');
+  andThen(() => {
+    assert.equal(find('.upp-low').text(), '1');
+  });
+  click('.btn-upp');
+  andThen(() => {
+    assert.equal(find('.upp-low').text(), '2');
+  });
+  click('.btn-upp');
+  andThen(() => {
+    // remains 2 because of logic in the action
+    assert.equal(find('.upp-low').text(), '2');
+  });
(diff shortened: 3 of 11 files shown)


tests/dummy/app/components/less-boilerplate/component.js
```diff
@@ -0,0 +1,28 @@
...
+
+var LessBoilerplateComponent = Ember.Component.extend({
+  actions: {
+    handleClick() {
+      this.actions['bumpTwice'](1);
+    }
+  },
+  layout: hbs`
+    <span class="upp-low">{{low}}</span>
+    <button class="btn-upp" onclick={{action "handleClick"}}>upp</button>
```
jkarsrud: Just curious, could we add another button here that actually passes in the initial value? As long as you don't need the actual jQuery event, won't this be a solution too; 

```hbs
<button onclick={{action "bumpTwice" 1}}>upp with value</button>
```

Unless I've completely forgotten something about the way actions curry their arguments in Ember though?

toranb: @jkarsrud good feedback - let me take another swing at this PR

toranb:
@jkarsrud @brettburley I've updated this to flip between thunk/saga on demand (required some require.unsee magic).

About the "extra button" comment - I decided to just do 1 button here for now as the purpose was to show the entire flow end-to-end and I think a single button achieves that.

The last question is regarding the lack of `this` (context) for action creators as @brettburley mentioned in the other open PR. Is this a show stopper? I assume people using action creators aren't tapping into the local component state but I could be wrong. Is it fair to call that a "nice to have" or is it a complete show stopper?

And finally - @jkarsrud could we close out your original PR and move forward with this? Any blockers not listed worth talking about before we merge this/ or consider merging it?

tests/helpers/start-app.js
```diff
@@ -28,7 +30,48 @@ export default function startApp(attrs) {
   let attributes = Ember.merge({}, config.APP);
   attributes = Ember.merge(attributes, attrs); // use defaults, but you can override;
 
+  require.unsee('dummy/services/redux');
```
brettburley: I think the reason this wasn't working originally was because the requires are not returned to default (i.e. cleaned up) properly after the test.

Could you do something like this instead?
https://github.com/brettburley/ember-redux/commit/1309ad6142362b41908e73b0c297d4190592ff64

I think that's a bit cleaner to keep this require swapping outside of the main application boot in tests, and keeps the swapping isolated to that particular test.

addon/components/connect.js
```diff
@@ -37,9 +38,16 @@ export default (stateToComputed, dispatchToActions=() => ({})) => {
...
-        );
+        if (typeof dispatchToActions === 'function') {
+          this.actions = Object.assign({},
+            this.actions, dispatchToActions.call(this, redux.dispatch.bind(redux))
+          );
+        }
+
+        if (typeof dispatchToActions === 'object') {
+          this.actions = Object.assign({},
+            this.actions, bindActionCreators.call(this, dispatchToActions, redux.dispatch.bind(redux)));
```
brettburley: Given that this doesn't appear to bind `this`, could you just change this to:
```
bindActionCreators(dispatchToActions, redux.dispatch.bind(redux));
```

That way it's not misleading to someone who might be reading this in the future.

brettburley:
Left a couple comments. This looks good to me, but if we can clean up that middleware patching for the saga test (see my other comment), that would be great :+1:

addon/components/connect.js
jkarsrud: I think that makes sense. That way the context of the action creator isn't the component calling it. If you need properties from your component, they should be passed explicitly.

addon/components/connect.js
```diff
@@ -37,9 +38,16 @@ export default (stateToComputed, dispatchToActions=() => ({})) => {
...
-        this.actions = Object.assign({},
-          this.actions, dispatchToActions.call(this, redux.dispatch.bind(redux))
-        );
+        if (typeof dispatchToActions === 'function') {
+          this.actions = Object.assign({},
+            this.actions, dispatchToActions.call(this, redux.dispatch.bind(redux))
+          );
+        }
+
+        if (typeof dispatchToActions === 'object') {
```
sandersky: I just want to note that `typeof [] === 'object'` evaluates to true, so you might actually be better off going with `Ember.typeOf(dispatchToActions) === 'object`. If you want to also support Ember.Object's then you could do something like `['instance', 'object'].indexOf(typeOf(dispatchToActions)) !== -1`.

jkarsrud: That's a good point. There is reason to what seems to be an error, though.

redux's `bindActionCreators` uses `Object.keys` to iterate over the object passed in to it. If given an array of functions, it will work just fine, as `Object.keys([() => 'func'])` will simply return `["0"]`, thus making it possible to look up by index from the original array, and do its thing. 

So while no-one actually might do it, we can actually use an object *or* an array as the value of `dispatchToActions`. It also gives us parity with the `react-redux` `connect` HOC (it only checks for `typeof mapDispatchToProps === 'object'`), so we get some docs for free 👍

toranb:
@brettburley thanks for the patch middleware revision - glad to have another set of eyes on this to show me a better way :)

toranb:
This feature is now included in the v2.3.0 version on npm

Issue #48, opened 2016-11-21
1.9 introduced regression in updateProps


dbashford:
This did not occur in `1.8`.

We've got a ton of components so it's a bit hard to track this down, but wanted to report what i saw in case it screams at anyone straight away.

Here's a bit of ember-redux code where I tossed a `console.log`

![image](https://cloud.githubusercontent.com/assets/581127/20498806/8d7b8352-affc-11e6-9dbf-9001b418c084.png)

This `_state` is an Ember thing, not a Redux thing.

Upon clicking an icon in our app, which triggers a state change, this is the output I see

![image](https://cloud.githubusercontent.com/assets/581127/20498870/d97cac54-affc-11e6-9094-0895220878f6.png)

249 components which report a `_state` as being `inDOM` and everything works fine.

1 component that reports a `_state` of `preRender` and then the __next line__ breaks.

22 components which report a `_state` of `undefined` which could be because of the breakage above it.

toranb:
@dbashford ouch - not good (sorry for the regression here)

In the short term - can you drop this into your node_modules and see if it's a quick win (no regression)? If yes we can patch this back tonight and learn more about what broke for you (curious myself)

(below was the connect source just before the big setProperties re-work)

https://github.com/toranb/ember-redux/blob/cfc1d2688034040fad1c15a520c79929d508c49d/addon/components/connect.js

dbashford:
You need me to do that to prove something out on your end? If not, I'm cool hanging at 1.8 for awhile. No rush.

toranb:
@dbashford mostly to confirm (for you) the 1 commit that did cause the break :) to help us narrow it down. I'd love if you could reproduce this with just a handful of components in a twiddle but didn't know if it was possible

dustinfarris:
@dbashford anything special about the components that are "_state === undefined"?

The `init` of the wrapped component sets an empty object for the `_props` property, so it's odd that it becomes `null` at some point in your application.  I wonder if _props is reserved elsewhere.  I'll dig into that.

dustinfarris:
@dbashford what version of ember are you on?

dbashford:
It's hard for me to ignore the `preRender` thing. That's the only piece of data that changes from component to component prior to the breakage besides those things you'd expect to change.  In the screenshot above you can see 249 components are `inDOM` but as soon as one is `preRender` things shut down.

This might be really hard to prove out in a twiddle. But I may be able to narrow what is going on in our app down a bit.

[...]

Issue #55, opened 2016-12-01
Need better error if trying to set property that already exists


dustinfarris:
e.g., this would error:

```hbs
{{my-component name="Something"}}
```

```js
const stateToComputed = () => {
  return {
    name: "Something else"
  }
};
```

Right now the error is `can't set read-only property "name" on object`.  We could probably trap this provide some guidance on what likely went wrong.

toranb:
@dustinfarris so the issue is the attr conflict - can you write up something to look for a conflict in the init/or didRecieveAttrs (not sure which is the best place w/out looking at a stacktrace myself)

I see this as a solid quality of life improvement @dustinfarris and others seem to agree :)

dustinfarris:
Yes, I'll take a swing at this

toranb:
@dustinfarris you the man! +100 !!!

toranb:
@brettburley fixed this in #117

PR #110, opened 2017-04-23
Only call stateToComputed once per state update.

brettburley:
We tried this a while ago in #41, but had to revert. It seems the revert was due to a collision in the `_props` key in #48 (@dustinfarris were you ever able to confirm this?).

Inspired by that issue, I'm proposing this variant which uses only a local variable `state` to maintain the state, which would prevent other addons/components/etc. from having a collision with the property since it's not defined on the component itself, and prevents folks using this addon from accessing the mapped state directly from within a component. Generally, it no longer exposes private methods like `getAttrs`, `handleChange`, nor private properties like `this.unsubscribe` to the consuming component.

### DIFF
--- a/addon/components/connect.js
+++ b/addon/components/connect.js
@@ -2,39 +2,103 @@ import Ember from 'ember';
 import { bindActionCreators } from 'redux';
 
 const {
+  assert,
   computed,
   defineProperty,
+  getProperties,
   inject: { service },
-  isEmpty,
   run
 } = Ember;
 
-export default (stateToComputed, dispatchToActions=() => ({})) => {
+/**
+  Returns a list of keys that have different values between the two objects.
 
-  if (!stateToComputed) {
-    stateToComputed = () => ({});
-  }
+  @method changedKeys
+  @return {Array} keys that have changed
+  @private
+*/
+function changedKeys(state, newState) {
+  return Object.keys(state).filter(key => {
+    return state[key] !== newState[key];
+  });
+}
 
+/**
+  Creates a read-only computed property for accessing redux state.
+
+  @method computedReduxProperty
+  @return {Function} an Ember computed property
+  @private
+*/
+function computedReduxProperty(key, getState) {
+  return computed({
+    get: () => getState()[key],
+    set: () => assert(`Cannot set redux property "${key}".`)
+  });
+}
+
+/**
+  Return an object of attrs passed to this Component.
+
+  `Component.attrs` is an object that can look like this:
+
+    {
+      myAttr: {
+        value: 'myValue'
+      }
+    }
+
+  Ember provides that a `get` will return the value:
+
+    this.get('myAttr') === 'myValue'
+
+  @method getAttrs
+  @return {Object} an object of key, value for each attr
+  @private
+*/
+function getAttrs(context) {
+  const attrKeys = Object.keys(context.attrs || {});
+  return getProperties(context, attrKeys);
+}
+
+export default (stateToComputed, dispatchToActions) => {
   return Component => {
-
     return Component.extend({
-
       redux: service(),
 
       init() {
         const redux = this.get('redux');
 
-        let props = stateToComputed.call(this, redux.getState(), this.getAttrs());
+        if (stateToComputed) {
+          const getState = () => stateToComputed.call(this, redux.getState(), getAttrs(this));
 
-        Object.keys(props).forEach(name => {
-          defineProperty(this, name, computed(() =>
-            stateToComputed.call(this, redux.getState(), this.getAttrs())[name]
-          ).property().readOnly());
-        });
+          let state = getState();
 
-        if (!isEmpty(Object.keys(props))) {
-          this.unsubscribe = redux.subscribe(() => {
-            this.handleChange();
+          Object.keys(state).forEach(key => {
+            defineProperty(this, key, computedReduxProperty(key, () => state));
+          });
+
+          const handleChange = () => {
+            const newState = getState();
+            if (newState === state) return;
+
+            const propsToNotify = changedKeys(state, newState);
+
+            state = newState;
+
+            if (propsToNotify.length > 0) {
+              run.join(() => {
+                propsToNotify.forEach(key => this.notifyPropertyChange(key));
+              });
+            }
+          };
+
+          const unsubscribe = redux.subscribe(handleChange);
+
+          this.on('didUpdateAttrs', handleChange);
+          this.one('willDestroyElement', () => {
+            this.off('didReceiveAttrs', handleChange);
+            unsubscribe();
           });
         }
 
@@ -50,57 +114,6 @@ export default (stateToComputed, dispatchToActions=() => ({})) => {
         }
 
         this._super(...arguments);
-      },
-
-      handleChange() {
-        const redux = this.get('redux');
-
-        const props = stateToComputed.call(this, redux.getState(), this.getAttrs());
-
-        const notifyProperties = Object.keys(props).filter(name => {
-          return this.get(name) !== props[name];
-        });
-
-        if (notifyProperties.length > 0) {
-          run.join(() => {
-            notifyProperties.forEach(name => this.notifyPropertyChange(name));
-          });
-        }
-      },
-
-      /**
-       * Return an object of attrs passed to this Component.
-       *
-       * `Component.attrs` is an object that can look like this:
-       *
-       *   {
-       *     myAttr: {
-       *       value: 'myValue'
-       *     }
-       *   }
-       *
-       * Ember provides that a `get` will return the value:
-       *
-       *   this.get('myAttr') === 'myValue'
-       *
-       * @method getAttrs
-       * @private
-       */
-      getAttrs() {
-        return this.getProperties(Object.keys(this.attrs || {}));
-      },
-
-      didUpdateAttrs() {
-        this._super(...arguments);
-        this.handleChange();
-      },
-
-      willDestroy() {
-        this._super(...arguments);
-        if (this.unsubscribe) {
-          this.unsubscribe();
-          this.unsubscribe = null;
-        }
       }
     });
   };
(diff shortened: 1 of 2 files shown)


addon/components/connect.js
```diff
@@ -1,98 +1,113 @@
...
+  Creates a read-only computed property for accessing redux state.
+
+  @method computedReduxProperty
+  @return {Function} an Ember computed property
+  @private
+*/
+function computedReduxProperty(key, getState) {
+  return computed({
+    get: () => getState()[key],
+    set: () => assert(`Cannot set redux property "${key}".`)
```
brettburley: While I was here, redefining the getter, I figured I might as well prevent `set`s with a more descriptive error per #55.

tests/integration/connect-test.js
```diff
@@ -71,6 +73,29 @@ test('stateToComputed can be used with component level CP if notifyPropertyChang
...
+    callCount++;
+    return { callCount };
+  }
+  this.register('component:test-component', connect(stateToComputed)(Component.extend({
+    layout: hbs`{{callCount}}`
+  })));
+
+  this.render(hbs`{{test-component attr=attr}}`);
+  assert.equal(this.$().text(), '1');
+  assert.equal(callCount, 1);
```
brettburley: This is the major change by calling `stateToComputed` only once. Prior to this change, this test reported `2`, `4`, and `6`. So each time the state or attrs changed, `stateToComputed` was invoked 2x as needed (once to figure out which properties had changed, and then once again for each property that is fetched during the `get` -- `callCount` in this case).

addon/components/connect.js
toranb: @dustinfarris will be so proud :)

addon/components/connect.js
```diff
@@ -1,98 +1,113 @@
...
+/**
+  Returns a list of keys that have different values between the two objects.
+
+  @method changedKeys
+  @return {Array} keys that have changed
+  @private
+*/
+function changedKeys(state, newState) {
+  return Object.keys(state).filter(key => {
+    return state[key] !== newState[key];
```
toranb: I assume this is a 1 for 1 replacement of the work we did previously in `handleChange` ?

```js
        const notifyProperties = Object.keys(props).filter(name => {
          return this.get(name) !== props[name];
        });
```

addon/components/connect.js
```diff
@@ -1,98 +1,113 @@
...
-        let props = stateToComputed.call(this, redux.getState(), this.getAttrs());
+        if (stateToComputed) {
+          const getState = () => stateToComputed.call(this, redux.getState(), getAttrs(this));
 
-        Object.keys(props).forEach(name => {
-          defineProperty(this, name, computed(() =>
-            stateToComputed.call(this, redux.getState(), this.getAttrs())[name]
-          ).property().readOnly());
-        });
+          let state = getState();
```
toranb: This is the bulk of the value add here correct? We cut down on the work required because you call `getState` just 1x now. The tradeoff here is that we are storing state locally in the component itself - any challenges/ or issues with this? I was afraid of the idea conceptually because although we called `getState` 2x we didn't need to worry about state synchronization (as each call for the state backing a CP would call out for it on demand)

Just curious what we are trading here if anything

addon/components/connect.js
```diff
@@ -1,98 +1,113 @@
...
-        });
+          const unsubscribe = redux.subscribe(handleChange);
 
-        if (notifyProperties.length > 0) {
-          run.join(() => {
-            notifyProperties.forEach(name => this.notifyPropertyChange(name));
+          this.on('didUpdateAttrs', handleChange);
+          this.one('willDestroyElement', () => {
+            this.off('didReceiveAttrs', handleChange);
+            unsubscribe();
```
toranb: What are the pros/cons of not exposing a separate `willDestoryElement` hook like we did previously? I feel like the `this.one` / `this.off` is a bit magical myself (assuming for the moment this isn't vanillaJS stuff here and it is only possible because of ember). Thoughts? How married are you to making this change now? pros/cons to making this change? For example: what if someone using the 2X had an override here (or inside the didUpdateAttrs) do they lose anything? Could this be considered a breaking api change? Was `getAttrs` a public api people could have tapp'd into? Or is that considered an intimate api?

tests/integration/connect-test.js
toranb: @brettburley love to see you adding this test back to the mix - I feel it's a good one to have regardless :)

tests/integration/connect-test.js
```diff
@@ -88,47 +113,35 @@ test('each computed is truly readonly', function(assert) {
     assert.throws(() => {
       this.$('.btn-alter').trigger('click');
     }, (e) => {
-      return e.message.indexOf('Cannot set read-only property') > -1;
+      return e.message.indexOf('Cannot set redux property "low"') > -1;
     });
   });
 });
 
 test('lifecycle hooks are still invoked', function(assert) {
```
toranb: This test even possible at the integration level (without adding additional monkey patching/hacks to confirm the lifecycle hooks are indeed invoked?) Example: if we had kept the willDestory/ didUpdateAttrs separate we would be able to see this more easily correct? One of the reasons I prefer those explicit hooks is that they allow me (the test driven developer) an entry point to do something (here for verification/ but you can imagine other scenarios I'm sure)

toranb:
@brettburley also - could you rebase this at some point :) my other PR conflicted it seems

addon/components/connect.js
brettburley: Yep, just pulled it out into a method to make it more descriptive.

addon/components/connect.js
brettburley: Yes, this is all the value here. I thought through the edge cases I can think of (particularly with observers, and computed properties based on redux properties), and cannot come up with any scenarios that might break. Given the synchronous nature of observers, and the fact that the existing properties are cached until `notifyPropertyChange` anyway, the behavior is almost identical.

addon/components/connect.js
brettburley: I'll start with the cons:
* Less explicit that we're relying on this hook.
* No control of the sequence with respect to the superclass invocation (`this._super`).

The pro is that by defining everything within our `init` function, we can take advantage of the closure and keep all of our state local and internal to `ember-redux`, without exposing any internal data (like the `state` property and the `unsubscribe` method).

There are two alternatives I can think of right now:
1. Keep our existing implementation where we define state on the component instance itself. In #41 @toranb tried this, but it seems like the `_props` key he used might have conflicted with another addon. We could use something more namedspaced like `__ember_redux_state__` as a key instead.
2. Use the component `guid` (from Ember) and store this state in a local, global hash from within ember-redux, like:
```
const subscriptionForComponent = {}
...
export default (stateToComputed, dispatchToActions) => {
  return Component => {
    return Component.extend({
      ...
      willDestroy() {
        this._super(...arguments);
        if (subscriptionForComponent[guidFor(this)]) {
          subscriptionForComponent[guidFor(this)]();
          delete subscriptionForComponent[guidFor(this)];
        }
      }
...
```

I actually rather like the second, but I'm not sure if there's any more general commentary or literature about keeping local state within an es module.

brettburley: > Could this be considered a breaking api change?
It could be, but I think given that these methods/features weren't documented, I doubt they were used much. I can't think of a case where you'd ever want to override `willDestroy` and not clean up the subscription. Nor can I think of any use case for overriding `handleUpdate`. However, there could be use cases for manually invoking `handleUpdate` yourself, but I'd be curious to find out if anyone is using that way, and understand the use case better.
> Was getAttrs a public api people could have tapp'd into? Or is that considered an intimate api?
It could be, but it's so generic that it would be trivial to re-implement, or if we felt it really was useful, we could re-expose it as a util function.

In general, I am a strong proponent for keeping a lot of these methods and data internal as to not expose them to the consuming component because it effectively becomes a mixin which introduces implicit dependencies that are unclear.

tests/integration/connect-test.js
brettburley: That's true, but for the same reason it makes them more testable, it also makes them more visible to a consuming app (which I think is generally bad, since we're exposing internals). However, I do think the value of the explicit definition (we're trained to look for these hooks now, and you can be specific about the order with the `this._super` call), is huge, and would love another way to keep the state internal to `ember-redux` while still using them.

brettburley:
Thanks for the review @toranb. Please see my comments inline. The rebase caused a few to be collapsed due to "outdated" -- but please take a look at those too!

addon/components/connect.js
brettburley: That second option was kind of interesting, so I sketched out what it might look like here: https://github.com/brettburley/ember-redux/commit/373b399e93702fc726086b781b69fc0dfc8c9404

brettburley:
@toranb I went ahead and pushed up another commit which brings back the original `didUpdateAttrs` and `willDestroy` hooks. I do prefer these hooks since the sequencing is more explicit, and it's been recommended to use these built-in hooks over the Ember.Evented interfaces. Also, I noticed that my original code was using `willDestroyElement` which is subtly different than the Object `willDestroy`, and I didn't want to introduce yet another change.

I am still maintaining keeping the `unsubscribe` and `handleChanges` methods internal (inspired by usages in other addons of `ember-weakmap` and similar). If you think that it's likely enough that others might have hooked into the `getAttrs` or `handleChanges` methods, I'd be glad to add aliases back for these methods with a deprecation warning if they're accessed.

toranb:
@brettburley I'd like to propose we shelve the work in this PR around `hooksForComponent` so we can focus in on the performance improvements you made here. This will also allow more community discussion centered around the public/private hooks decision specifically. Would you be open to a new PR after we merge this one that adds the hooks rework?

Note: not saying we won't land it btw - just think this PR could be solely focused on the performance work (as I don't see any reason we couldn't merge that today/ and move ahead)

brettburley:
Opened #114.

Issue #112, opened 2017-04-26
RFC: Export route and connect from addon root.


brettburley:
# Summary
The addon root (`addon/index.js`) should export the public API methods for `ember-redux`. The two public methods, `route` and `connect` would be exported so that they could be used in a consuming app via:
```
import { route, connect } from 'ember-redux';
```

# Motivation
Currently the `route` and `connect` functions are exported as default functions from `ember-redux/route` and `ember-redux/components/connect`. The connect function is not really a component, so importing it from a components directory is a bit of a misnomer. More importantly, many other popular ember addons have standardized on exporting their main APIs from their root (e.g. `ember-concurrency`, `ember-data`).

This makes the imports short and memorable, and clarifies which files and methods are part of the public api.

# Detailed Design
We would move the `connect.js` file into `addon` to live alongside the existing `route.js`. The directory structure would look like:
```
addon/index.js
addon/connect.js
addon/route.js
```
The `addon/index.js` file would simply import and re-export the `connect` and `route` files. 

Additionally, we can maintain a file at `addon/components/connect.js` which would import from `addon/connect.js` and re-export with a console deprecation message directing folks using the addon to update their imports. I expect updating the import statements should be very straightforward, and should be possible in a bulk find & replace in almost all cases. At some future version, we can completely remove the old `addon/components/connect.js` file completely.

# Alternatives
Don't make any changes and leave the files where they are.

toranb:
@brettburley great write up - no objections here. I assume we will move ahead w/ your PR next week if we don't get any push back

toranb:
@brettburley well it's been a solid month and then some - the final decision looks/feels like a yes on this. I'll close this out and we can move forward in that PR you have up

PR #113, opened 2017-04-26
Export route and connect from addon root.

brettburley:
This is an example implementation for #112.


addon/connect.js
```diff
@@ -0,0 +1,107 @@
+import Ember from 'ember';
```
brettburley: This is just a move (no changes) from its original location.

toranb:
@brettburley can't say I have a problem with this - seems to make sense. Anyone else pros/cons to making this change (aside from the find/replace for import statements when we land a v3.0.0 later this year) ?

toranb:
@brettburley I went ahead and rebased this branch - it seems g2g but I wanted your final input before we roll ahead. On a related note, this made the typescript definition work much easier so that is a large reason I'm suddenly motivated to land this ;)

brettburley:
Yep, this still looks good to me!

toranb:
@brettburley this is now published in ember-redux v2.7.0 (I also updated the guides to reflect this new import location)

PR #114, opened 2017-04-26
Only call stateToComputed once per state update (v2).

brettburley:
This is a stripped down version of #110, addressing only the stateToComputed issue. I'll open separate PRs for the other pieces of that PR coming soon.

### DIFF
--- a/addon/components/connect.js
+++ b/addon/components/connect.js
@@ -5,36 +5,46 @@ const {
   computed,
   defineProperty,
   inject: { service },
-  isEmpty,
   run
 } = Ember;
 
 export default (stateToComputed, dispatchToActions=() => ({})) => {
-
-  if (!stateToComputed) {
-    stateToComputed = () => ({});
-  }
-
   return Component => {
-
     return Component.extend({
-
       redux: service(),
 
       init() {
         const redux = this.get('redux');
 
-        let props = stateToComputed.call(this, redux.getState(), this.getAttrs());
+        if (stateToComputed) {
+          const getProps = () => stateToComputed.call(this, redux.getState(), this.getAttrs());
+          let props = getProps();
 
-        Object.keys(props).forEach(name => {
-          defineProperty(this, name, computed(() =>
-            stateToComputed.call(this, redux.getState(), this.getAttrs())[name]
-          ).property().readOnly());
-        });
+          Object.keys(props).forEach(name => {
+            defineProperty(this, name, computed(() =>
+              props[name]
+            ).property().readOnly());
+          });
 
-        if (!isEmpty(Object.keys(props))) {
+          this._handleChange = () => {
+            const newProps = getProps();
+            if (props === newProps) return;
+
+            const notifyProperties = Object.keys(props).filter(name => {
+              return props[name] !== newProps[name];
+            });
+
+            props = newProps;
+
+            if (notifyProperties.length > 0) {
+              run.join(() => {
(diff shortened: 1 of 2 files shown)


addon/components/connect.js
```diff
@@ -53,18 +63,14 @@ export default (stateToComputed, dispatchToActions=() => ({})) => {
...
-
-        const notifyProperties = Object.keys(props).filter(name => {
-          return this.get(name) !== props[name];
-        });
-
-        if (notifyProperties.length > 0) {
-          run.join(() => {
-            notifyProperties.forEach(name => this.notifyPropertyChange(name));
-          });
+        Ember.warn(
```
brettburley: There's a subtle change here in that consumers who were overriding this method and calling the original with super (e.g. if they wanted to do more work here) will not have that ability anymore. I doubt anyone would have been doing that, because you would have to write something like:

```
export default connect(stateToComputed, dispatchToActions)(Ember.Component.extend({
...
})).extend({
  handleChange(...args) {
    // Your custom code here...
    this._super(...args);
  }
});
```

toranb: Right on - I think you've done the right thing here moving us forward while warning people of a soon to be dead api :)

addon/components/connect.js
```diff
@@ -5,36 +5,46 @@ const {
...
+
+          Object.keys(props).forEach(name => {
+            defineProperty(this, name, computed(() =>
+              props[name]
+            ).property().readOnly());
+          });
+
+          this._handleChange = () => {
+            const newProps = getProps();
+            if (props === newProps) return;
```
toranb: This is the only line I don't see adding value visibly (if I console log when this condition is true I never see anything showing this is useful). That said, I do understand it conceptually (value based equality with immutability). Are you able to see this saving cpu cycles down the road on big apps?

addon/components/connect.js
```diff
@@ -92,7 +98,11 @@ export default (stateToComputed, dispatchToActions=() => ({})) => {
 
       didUpdateAttrs() {
         this._super(...arguments);
-        this.handleChange();
+        // Components only have a handleChange method if it is subscribed
+        // to redux changes.
+        if (this._handleChange) {
```
toranb: If you don't mind I'll remove these 2 comments as they are self evident above (reading the source for 10 min will expose why you added this conditional)

tests/integration/connect-test.js
```diff
@@ -173,3 +198,23 @@ test('connecting dispatchToActions as object should dispatch action', function(a
     this.$('.btn-down').trigger('click');
   });
 });
+
+test('calling handleChange directly shows a deprecation warning', function(assert) {
+  assert.expect(1);
```
toranb: great to have a test for this - likely we will cut it at v3 later this year

toranb:
@brettburley @dbashford this is now up on npm as v2.4.0

Could you guys test this out at some point this week/ early next? It's a fairly minimal change on the surface but I want to be sure all is good for those of you with real apps in the wild :)

addon/components/connect.js
brettburley: Yeah, so this is only useful with things like reselect. Since it memoizes the output of the stateToComputed method, if the state inputs are the same, it'll return the same props object, which means we don't have to do any processing on it!

addon/components/connect.js
brettburley: Yeah sure thing.

PR #116, opened 2017-04-27
Clean up functions used in connect.

brettburley:
This is some of the work I had bundled in #110. This makes `getAttrs` a private (static) function with a deprecation warning. I also pulled out the `changedKeys` utility function to be a bit more descriptive.


toranb:
@brettburley if you can rebase this, I'll pull it in tonight/tomorrow :) 

thanks again for pushing ahead with this recent block of improvements!

PR #117, opened 2017-04-27
Add a more descriptive error message on redux prop set.

brettburley:
Per #55.


toranb:
@brettburley I think this one is fairly straight forward :) thanks for making a more useful error message!

PR #118, opened 2017-04-28
[TESTS]: added queryParams + connect example

toranb:
This is the bulk of @sandersky 's great work in a older PR he wrote. I think it's a great example I can point people at and because it was so well thought out I decided to include it in the project :)


sandersky:
:+1:

Issue #115, opened 2017-04-27
Question -- how combine reducers in ember-redux?


skitterm:
I saw examples of combining reducers over in the redux.js.org [docs](http://redux.js.org/docs/basics/Reducers.html#splitting-reducers). I was wondering how to go about this in Ember redux? For this example, I have a `settings` and a `users` reducer, that will manage those pieces of the state respectively, so the store's state shape should be
```javascript
{
  settings: // settings data,
  users: // users data
}
```

I tried bringing in redux and doing this with `combineReducers`, but I got an error -- here are my reducers:
```javascript
// app/reducers/settings.js
const settings = (state, action) => {
  // reducer code here
};

export default {
  settings
}


// app/reducers/users.js
import { combineReducers } from 'redux';
import settings from './settings';
const users = (state, action) => {
  // reducer code here
};

export default combineReducers({
  settings,
  users
})
```
It worked if I switched the exports for the two reducers to be as follows:
```javascript

// settings.js
export default settings

// users.js
export default {
  settings,
  users
}
```

But I wasn't sure if that does what it needs to. The redux docs show a longhand `combineReducers` equivalent like this:
```javascript
// users.js
function rootReducer(state = {}, action) {
  return {
    settings: settings(state.settings, action),
    users: users(state.users, action)
  };
}

export default rootReducer
```
But that didn't work for me either.

What is the `ember-redux` way to combine reducers together to compose a state tree?

toranb:
@skitterm great question! The short version is to add a reducers directory and inside of it an index.js file that simply exports each independent reducer you have in your app/ or addon(s).

For example, In this example app the index.js file pulls in each reducer and simply exports it. 

https://github.com/toranb/ember-addon-redux-style/blob/master/yelp/app/reducers/index.js

Then, inside ember-redux (the service) -we call `combineReducers` using this index.js exported value.

https://github.com/ember-redux/ember-redux/blob/master/app/services/redux.js#L26

We don't layer in a lot of convention compared to vanilla redux but this is one case in which we simply chose to group the reducers together (under 1 directory) and export them as a single dict. This also allowed us to provide a configuration-free store that is wired in the default redux service used by "connected" components (the ones that subscribe to data in the redux store itself).

Any other questions let me know - happy to help :)

skitterm:
[...]

PR #119, opened 2017-04-29
[FEATURE]: allow root reducer to explicitly call combineReducers from…

toranb:
In issue #115 it was requested that we allow `combineReducers` from the root reducers/index 

This shows that working end-to-end w/out breaking people in the 2X series today (who are not using `combineReducers` directly)

### DIFF
--- a/tests/dummy/app/reducers/combined.js
+++ b/tests/dummy/app/reducers/combined.js
@@ -1,0 +1,10 @@
+import Ember from 'ember';
+
+var combined = function combined(state, action) {
+  if (action.type === 'COMBINE') {
+    Ember.assert('the reducers patch failed or was not applied');
+  }
+  return state || { done: false };
+};
+
+export default combined;
--- a/tests/helpers/patch-reducer.js
+++ b/tests/helpers/patch-reducer.js
@@ -1,0 +1,34 @@
+/* global require, define */
+
+import originalReducer from 'dummy/reducers/index';
+
+const { unsee } = require;
+
+export function applyPatch() {
+  unsee('dummy/services/redux');
+  unsee('dummy/reducers/index');
+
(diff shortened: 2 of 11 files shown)


tests/dummy/app/reducers/combined.js
```diff
@@ -0,0 +1,10 @@
+var combined = function combined(state, action) {
+  if (action.type === 'COMBINE') {
+    // so we fail the patch reducers test if the
+    // patch fails/ or didn't apply as expected
+    // return Object.assign({}, state, { done: true });
```
brettburley: It might be better to throw an explicit `Ember.assert` in here if this is invoked to make it clearer to someone reading this code in the future.

tests/helpers/patch-reducer.js
```diff
@@ -0,0 +1,41 @@
...
+const { unsee } = require;
+
+export function applyPatch() {
+  unsee('dummy/services/redux');
+  unsee('dummy/reducers/index');
+
+  define('dummy/reducers/index', ['exports', 'dummy/reducers/low', 'dummy/reducers/high', 'redux'], function (exports, _low, _high, _redux) {
+    'use strict';
+
+    Object.defineProperty(exports, "__esModule", {
```
brettburley: Did you find the need for this `__esModule` definition? I think it's not required with the loader ember-cli uses, but might be mistaken on best practices.

tests/dummy/app/reducers/combined.js
toranb: good call ! will update this

tests/helpers/patch-reducer.js
toranb: ah yes, I'll emulate what the middleware patch looks like

toranb:
@brettburley updated the source based on your comments just now. I'll wait to merge this until we hear from the original author of the issue but I think we can call this g2g :)

cc @skitterm

skitterm:
Tried using combineReducers at the root level, and it worked great! Also worked when I was just exporting a plain object.

PR #120, opened 2017-04-29
Migrate the redux services/reducers/middleware into the addon directory

toranb:
In issue #115 @foxnewsnetwork asked if we could migrate the redux service itself into the addon directory (making it easier to custom/override on demand).

This is that PR if people have feedback

note: the dummy app introduces some interesting pain points so I've had to hack around in the test-helper to work around this for now. I've got 2 skipped tests that I can't seem to get just yet - pushing this up for early feedback. The 2 failing (skipped) tests are just an artifact of the complex build reducer/middleware patching we do to exercise some customization requirements.

I'll work through those before we actually merge this - feedback welcome otherwise


toranb:
@foxnewsnetwork this also solves the issue people have booting ember-redux for the first time because we supply a default reducer that should be provided by default (from the end user). This is our number 1 pain point for new users/ on boarding so it could be a win/win

toranb:
cc @dbashford @brettburley 

This shouldn't be a breaking change but I do remember talking with @dbashford about how he is overriding middleware (or a reducer) in a dummy app and I'd like to have some confirmation that if we merged this it wouldn't ripple out and break the world :)

No huge rush - in the next 2 weeks if you both have time to help test out this branch :) I'd appreciate it

toranb:
For anyone reviewing - ignore my comments above about 2 failing tests (just fixed those up and force pushed away my regrets).

addon/middleware/index.js
```diff
@@ -0,0 +1,5 @@
+import thunk from 'redux-thunk';
```
foxnewsnetwork: Whoa!! This just works out of the box?! I was under the impression redux-thunk refused to be imported in `addon`. But if it works, then great (it also means I've lied in my PR to the thunk shim repo https://github.com/ember-redux/ember-redux-thunk-shim/pull/19)

app/services/redux.js
```diff
@@ -1,39 +1 @@
...
-    return this.store.getState();
-  },
-  dispatch(action) {
-    return this.store.dispatch(action);
-  },
-  subscribe(func) {
-    return this.store.subscribe(func);
-  }
-});
+export {default} from 'ember-redux/services/redux';
```
foxnewsnetwork: We might want to do something like:

```javascript
import middlewareConfig from '../middleware/index';
import ReduxService from 'ember-redux/services/redux';

export default ReduxService.extend({
  init() {
    makeStoreHere()
  }
})
```

Primarily because currently ember-redux pulls an end-user's reducers out of their `app/reducers/index` (same with enhancers and middleware), but if we just pull `service/redux` into addon, we'd lose that functionality

foxnewsnetwork: In other words, we'll still need to support pull in all the user's reducers, middlewares, and enhancers in the `app` directory's redux service, which probably means we might need to refactor our `addon/services/redux` a bit to be more accommodating to this.

Consider my implementation here: https://github.com/foxnewsnetwork/ember-redux/blob/feature/addonify-redux-service/addon/services/redux.js#L30

```javascript
import reducers from '../reducers/optional';
import middlewareConfig from '../middleware/index';

export default Ember.Service.extend({
  middlewareConfig,
  reducers,
  enhancers,
  makeStoreInstance,
  init() {
    // gets the middlewareConfig from the instance of this service
    // if you'd like to customize your middlewares, you'd probably do it here
    const middlewareConfig = get(this, 'middlewareConfig');
    const reducers = get(this, 'reducers');
    const enhancers = get(this, 'enhancers');

    this.store = this.makeStoreInstance({ middlewareConfig, reducers, enhancers });
    this._super(...arguments);
  }
...otherstuff
}
``` 

which would allow us in `app/services/redux` to just do:
```javascript
import ReduxService from 'ember-redux/services/redux';
import middlewareConfig from '../middleware/index';

export default ReduxService.extend({ reducers, enhancers, middlewareConfig });
```
https://github.com/foxnewsnetwork/ember-redux/blob/feature/addonify-redux-service/app/services/redux.js

foxnewsnetwork:
Let some review notes; also worked on my own branch for this feature before I realized you already have one opened (lol), linking here for reference sake: https://github.com/foxnewsnetwork/ember-redux/tree/feature/addonify-redux-service

toranb:
screen shot 2017-04-29 at 9 20 42 pm

My original comments about not having the ability to "pull this into addon" do seem correct now that I've actually installed it in a real app.

I'll look into your comments above and see what I can refactor toward. One thing to note (I'll mention it here and in your thunk PR) is that we can't require ember-cli v2.10+ so the move to `ember-cli-node-assets` simply isn't on the table right now. I know for a fact that a few end users are in a huge legacy code base and stuck on ember-cli v2.4.3.

That said, I still think we can make the "createStore" extensible while providing a sensible default for those who don't care about it.

toranb:
@foxnewsnetwork @dbashford @brettburley closing this PR out as it doesn't work and I've opened PR #121 in it's place

PR #121, opened 2017-04-30
A more robust createStore inspired by @foxnewsnetwork

toranb:
As part of the feedback for PR #120 @foxnewsnetwork passed over a createStore that is more extensible (while not breaking the app itself). I took inspiration from that and my previous PR to build this one #feedbackWelcome

### DIFF
--- a/addon/middleware/index.js
+++ b/addon/middleware/index.js
@@ -1,0 +1,1 @@
+export default [];
--- a/addon/reducers/index.js
+++ b/addon/reducers/index.js
@@ -1,0 +1,17 @@
+// please write your own index.js that exports legit reducers
+//
+// import { combineReducers } from 'redux';
+// const foobar = (state, action) => {
+//   //reducer code here
+// };
+//
+// export default combineReducers({
+//   foobar
+// })
+import Ember from 'ember';
+
+export default function(state) {
+  Ember.warn(`You haven't created a reducer yet. Place the root reducer in "app/reducers/index.js"`, null, { id: 'ember-redux.default-reducer' });
+
+  return state;
+}
--- a/addon/services/redux.js
+++ b/addon/services/redux.js
@@ -1,0 +1,53 @@
+import Ember from 'ember';
+import redux from 'redux';
+import reducers from '../reducers/index';
+import enhancers from '../enhancers/index';
+import middlewares from '../middleware/index';
+
+const { assert, isArray, get } = Ember;
+
+// Handle "classic" middleware exports (i.e. an array), as well as the hash option
+const extractMiddlewareConfig = (mc) => {
+  assert(
+    'Middleware must either be an array, or a hash containing a `middleware` property',
+    isArray(mc) || mc.middleware
+  );
+  return isArray(mc) ? { middleware: mc } : mc;
+}
+
+const { createStore, applyMiddleware, combineReducers, compose } = redux;
+
+const makeStoreInstance = ({middlewares, reducers, enhancers}) => {
+  const { middleware, setup = () => {} } = extractMiddlewareConfig(middlewares);
+  const createStoreWithMiddleware = compose(applyMiddleware(...middleware), enhancers)(createStore);
+  const reducer = typeof reducers === 'function' ? reducers : combineReducers(reducers);
+  const store = createStoreWithMiddleware(reducer);
+  setup(store);
+  return store;
+};
+
+export default Ember.Service.extend({
+  middlewares,
+  reducers,
+  enhancers,
+  makeStoreInstance,
+  init() {
+    const enhancers = get(this, 'enhancers');
+    const reducers = get(this, 'reducers');
+    const middlewares = get(this, 'middlewares');
+    this.store = this.makeStoreInstance({ middlewares, reducers, enhancers });
+    this._super(...arguments);
+  },
+  getState() {
+    return this.store.getState();
+  },
+  dispatch(action) {
+    return this.store.dispatch(action);
+  },
+  subscribe(func) {
+    return this.store.subscribe(func);
+  },
+  replaceReducer(nextReducer) {
+    return this.store.replaceReducer(nextReducer);
+  }
+});
--- a/app/reducers/index.js
+++ b/app/reducers/index.js
@@ -1,0 +1,1 @@
+export {default} from 'ember-redux/reducers/index';
--- a/app/reducers/optional.js
+++ b/app/reducers/optional.js
@@ -1,7 +1,0 @@
-//this is a simple pass thru that offers an extension point
-
-export default function(combine) {
-  return (state, action) => {
-    return combine(state, action);
-  };
-};
--- a/app/enhancers/index.js
+++ b/app/enhancers/index.js
@@ -1,1 +1,1 @@
-export {default} from 'ember-redux/enhancers'
+export {default} from 'ember-redux/enhancers/index';
--- a/app/services/redux.js
+++ b/app/services/redux.js
@@ -1,43 +1,6 @@
-import Ember from 'ember';
-import redux from 'redux';
+import ReduxService from 'ember-redux/services/redux';
 import reducers from '../reducers/index';
 import enhancers from '../enhancers/index';
-import middlewareConfig from '../middleware/index';
+import middlewares from '../middleware/index';
 
-const { assert, isArray } = Ember;
-
-// Handle "classic" middleware exports (i.e. an array), as well as the hash option
-const extractMiddlewareConfig = (mc) => {
-  assert(
-    'Middleware must either be an array, or a hash containing a `middleware` property',
-    isArray(mc) || mc.middleware
-  );
-  return isArray(mc) ? { middleware: mc } : mc;
-}
-
-// Destructure the middleware array and the setup thunk into two different variables
-const { middleware, setup = () => {} } = extractMiddlewareConfig(middlewareConfig);
-
-var { createStore, applyMiddleware, combineReducers, compose } = redux;
-var createStoreWithMiddleware = compose(applyMiddleware(...middleware), enhancers)(createStore);
-
-export default Ember.Service.extend({
-  init() {
-    const reducer = typeof reducers === 'function' ? reducers : combineReducers(reducers);
-    this.store = createStoreWithMiddleware(reducer);
-    setup(this.store);
-    this._super(...arguments);
-  },
-  getState() {
-    return this.store.getState();
-  },
-  dispatch(action) {
-    return this.store.dispatch(action);
-  },
-  subscribe(func) {
-    return this.store.subscribe(func);
-  },
-  replaceReducer(nextReducer) {
-    return this.store.replaceReducer(nextReducer);
-  }
-});
+export default ReduxService.extend({ reducers, enhancers, middlewares });
--- a/docs/configure.md
+++ b/docs/configure.md
@@ -4,6 +4,38 @@
 
 If you have a solid understanding of redux basics and what the ember bindings offer, this section will help you unlock the potential of ember-redux by showing all the extension points!
 
+**Customize createStore**
+
+If you need to customize how the redux store is created you can override the existing ember service. In the example below we import the reducers and enhancers like usual but notice we implement a method named `makeStoreInstance` and provide this to the new ReduxService we export.
+
+```js
+//app/services/redux.js
+import redux from 'redux';
+import ReduxService from 'ember-redux/services/redux';
+import createSagaMiddleWare from 'redux-saga';
+import reducers from '../reducers/index';
+import enhancers from '../enhancers/index';
+import root from '../sagas/index';
+
+const { createStore, applyMiddleware, compose } = redux;
+
+const sagaMiddleware = createSagaMiddleWare();
+
+const makeStoreInstance = ({reducers, enhancers}) => {
+  const middleware = applyMiddleware(sagaMiddleware);
+  const createStoreWithMiddleware = compose(middleware, enhancers)(createStore);
+  const store = createStoreWithMiddleware(reducers);
+  sagaMiddleware.run(root);
+  return store;
+};
+
+export default ReduxService.extend({
+  reducers,
+  enhancers,
+  makeStoreInstance
+});
+```
+
 **Enhancers**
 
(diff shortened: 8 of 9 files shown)


toranb:
screen shot 2017-04-29 at 9 51 14 pm

Proof this implementation not only has a green build but won't bust your app 👍

app/reducers/optional.js
```diff
@@ -1,7 +0,0 @@
-//this is a simple pass thru that offers an extension point
```
toranb: just for reference -this was removed from ember-redux just before the 2.0 release

https://github.com/ember-redux/ember-redux/commit/be2f989cea73a802a0afaacb011dffa59cef4e09

toranb:
I added a new commit to move the `redux-thunk` imports to app/ instead of addon/

By doing this the consuming app can truly/safely remove `redux-thunk` as a dependency and add something like `redux-saga` :)

toranb:
Because this was a move the service diff isn't a great read - if you want to see a simple diff I pushed a patch up to [dropbox](https://dl.dropboxusercontent.com/u/716525/downloads/opensource/theDiff.patch) (worth a look if you want a simple idea of what has changed in the service)

toranb:
I think the last remaining step here could/would? be to make `makeStoreInstance` something we import explicitly so "power users" can truly just override that part producing a store configured any way they want (like you would by hand using `react-redux` for example).

Anyone have thoughts on this? where would we export it from exactly?

foxnewsnetwork:
`makeStoreInstance` is already merged in with the other stuff as a part of extending `Ember.Service`

Therefore, power-users can already do something like:

```javascript
import ReduxService from 'ember-redux/services/redux';

export default ReduxService.extend({
  makeStoreInstance({ middlewareConfig, enhaners, reducers }) { 
    doWhatever()
  })
```

foxnewsnetwork:
LGTM. Also, please consider also editing over some of the howto extend from `README.markdown` stuff on my branch: https://github.com/foxnewsnetwork/ember-redux/tree/feature/addonify-redux-service#customizing-the-redux-service

addon/middleware/index.js
```diff
@@ -0,0 +1,13 @@
+//only here as a vanilla pass thru so we can
```
brettburley: What is this middleware for? Is it used at all?

addon/services/redux.js
```diff
@@ -0,0 +1,50 @@
+import Ember from 'ember';
+import redux from 'redux';
+import reducers from '../reducers/index';
```
brettburley: Since these imports are within the addon itself, should we actually exclude them here, and just rely on them being passed into the extended version of the service? I don't think they add any value here since we want to include the `reducers`, `enhancers`, and `middlewareConfig` from the host app anyway.

app/reducers/index.js
```diff
@@ -0,0 +1 @@
+export {default} from 'ember-redux/reducers/index';
```
brettburley: Previously, did someone using `ember-redux` always have to create a `reducers/index.js` file in order to prevent their app from breaking?

addon/services/redux.js
```diff
@@ -0,0 +1,50 @@
...
+  );
+  return isArray(mc) ? { middleware: mc } : mc;
+}
+
+const { createStore, applyMiddleware, combineReducers, compose } = redux;
+
+const makeStoreInstance = ({middlewareConfig, reducers, enhancers}) => {
+  const { middleware, setup = () => {} } = extractMiddlewareConfig(middlewareConfig);
+  const createStoreWithMiddleware = compose(applyMiddleware(...middleware), enhancers)(createStore);
+  const reducer = typeof reducers === 'function' ? reducers : combineReducers(reducers);
```
brettburley: Is this intentionally including the change from #119?

app/reducers/index.js
toranb: yes - it would throw an error letting you know the reducers/index.js was missing (not great but it also was a solid entry point into "starting using redux ..."). This improves it as we don't break out of the box (when this PR lands)

addon/services/redux.js
toranb: only because I'm 99% sure I'll merge that PR in the next 24 hours and I'll rebase this one right after (sorry for the confusion here - mostly trying to work conflict free)

addon/services/redux.js
toranb: @foxnewsnetwork ^^ any reason we can't/or shouldn't do this? on the surface it seems redundant as @brettburley mentioned

addon/middleware/index.js
toranb: because the addon/services/redux file calls out an explicit "import" for middleware I need this file present. If I include the `redux-thunk` version here (instead of app) the end user is then required to have redux-thunk installed even if they opt-out and use redux- saga instead

brettburley: Actually this file is an artifact of https://github.com/ember-redux/ember-redux/pull/121#discussion_r114086961, right?

Also, depending on the resolution of that comment, could this file be simply an `export default [];` or does that not work/am I missing something else?

toranb:
@brettburley I've rebased this and cut the addon/middleware down to just an empty array as you suggested. I still need to pull in some additional documentation for this as @foxnewsnetwork suggested above ^^

addon/reducers/index.js
```diff
@@ -0,0 +1,16 @@
...
+// const foobar = (state, action) => {
+//   //reducer code here
+// };
+//
+// export default combineReducers({
+//   foobar
+// })
+
+export default function(combine) {
+  return (state, action) => {
```
toranb: @brettburley should I add some type of console log or something here to alert the consuming app that they need to write a custom reducer?

addon/services/redux.js
toranb: @brettburley I've rebased this and it should be down to a single commit now :)

addon/reducers/index.js
brettburley: That seems like a reasonable approach. I've been using `Ember.warn` for those purposes (that way, it's stripped from production builds).

Also, doesn't this reducer completely fail as-is, since there's no combine being passed in? I think this file needs to just export a reducer, like:
```
export default function (state, action) {
 return state;
};
```

toranb:
Okay I've got this rebased - still needs some docs love (copy/paste of what Thomas linked us to above). I need to see how this looks/feels in a real app so that's on me this coming week

bstro:
Any estimate on when this sweet PR might land? Anything I can do to help?

toranb:
@bstro The only real blocker at the moment is my internal debate over the "breaking change" this introduced. The problem is that previously we wouldn't include a reducer for you (meaning an error message would pop up when you ran the build w/out `reducers/index.js`

We are planning to cut v3.0 in early July and this is likely the 2nd PR that will land officially (in that case any of the fears about a breaking api change will have a major version number to land under). 

I don't have anything else planned until that release next month but we could use some examples/ docs/ for people who want to extend this like yourself. Would you want to write up a small section for the docs that covers this in more detail?

bstro:
Ah, so this won't land in v2.x? In order to have these changes, I'd also need to upgrade our version of node+ember+ember-cli?

```
ember-cli: 2.4.3
node: 4.8.2
os: darwin x64
```

toranb:
@bstro It's possible we could land this in 2x -worth a debate on the "breaking change" I mentioned above. That said, in the discussion about 3X I showed a graph after node 8 was released recently and asked about what may be blocking you from taking on node 6 as a build time dependency.

You can still use ember-cli v2.4.3 w/ node 6.10 and that will be supported in both 2X and 3X of `ember-redux`. The problem arises when you need to build your ember-cli v2.4.3 app with node 4 (you will find a break as a result of the upgrade from babel 5 to babel 6).

What is blocking you/ your team from adopting node 6 as a build time dependency today?

brettburley:
@toranb I don't think it's a _breaking_ change since there aren't any meaningful apps with a missing `reducers/index.js` file. I don't think this PR changes the API in any way for existing apps, and only changes behaviors for new apps, so wouldn't necessitate a major version change because there is no work an app owner would have to do. I also think adding the runtime warning is probably sufficient warning for the new app case.

I'd strongly support landing this in the current major version (with the caveats of my previous feedback). :+1:

addon/services/redux.js
```diff
@@ -0,0 +1,53 @@
...
+  const store = createStoreWithMiddleware(reducer);
+  setup(store);
+  return store;
+};
+
+export default Ember.Service.extend({
+  middlewareConfig,
```
brettburley: Is there a reason this field is named `middlewareConfig` rather than just `middleware` (matching the pattern of `reducers`, `enhancers`)? Just want to be careful with the naming since these will become part of the public API when overriding the default implementation here.

bstro:
@toranb I'm going to spike this week on upgrading our Node to 6 LTS. Some promising progress so far.

toranb:
@brettburley it sounds like we agree that this change does not introduce a real/ functional breaking change of any kind. 

@brettburley I've updated the name you called out `middlewareConfig` to instead match the others (plural) `middlewares`. One question with that name change - is it correct? Meaning, we could bolt on 2+ middleware (one for logging/ one for async work/ etc)

The last step before we merge this into the 2x series is for me to find a way to test this from the dummy app (like we did w/ patchMiddleware). If that's not possible/or easily possible I'd like to test this manually with ember-cli v2.4.3 and v2.13.2

addon/services/redux.js
foxnewsnetwork: IIRC importing them there was for backwards compatibility reasons so that users of older version of ember-redux wouldn't have to change up the way they're importing... but if we're bumping major versions for these sets of changes, I think @brettburley 's approach is valid.

toranb:
@foxnewsnetwork @brettburley @bstro 

I threw together a working ember-cli v2.4.3 app using this fork and customized the redux store creation. Looking for feedback on this before we merge it next week

https://github.com/toranb/configme243/commit/647145fa679c58598a352bc6a25a9b93607270d8

bstro:
Your 2.4.3 example looks consistent with how I've been using it! It feels pretty good to me. Nothing to critique on my end.

toranb:
@bstro one issue I've found today is that I can destructure enhancers just fine but not the reducers for some reason. The below doesn't blow up, but the reducers don't fire as they do when I wire it up myself. I'll poke around at this until I can get a config that truly provides reducers, middlewares, and enchancers.

```js
export default ReduxService.extend({
  makeStoreInstance({reducers, enhancers}) {
    const middleware = applyMiddleware(sagaMiddleware);
    const createStoreWithMiddleware = compose(middleware, enhancers)(createStore);
    const store = createStoreWithMiddleware(reducers);
    sagaMiddleware.run(root);
    return store;
  }
});
```

bstro:
@toranb now that you mention it, I recall running into that too. I had to manually import my reducers.

toranb:
@bstro I believe the variables we destructure in `makeStoreInstance` reference the "default" imports that ship with `ember-redux` itself (meaning the reducer/middleware/enchancer we ship that do next to nothing).

@foxnewsnetwork is this how you'd expect this to work? Not a complete deal breaker obviously, just trying to understand how those vars coming into `makeStoreInstance` should work/ what the developer experience is like using them

bstro:
@toranb Doesn't that make the params to makeStoreInstance effectively useless?

bstro:
@toranb I don't think its a big deal to manually import reducers and enhancers...

toranb:
@bstro I'm personally good with this myself (just wasn't sure if that was the expectation when I took my first stab at extending it). This does have the added benefit of "no breaking change" for people in the 2X series. Plus it still allows us to provide a default middleware using `redux-thunk` without the user even knowing it was required (often the very first question people new to redux ask after "hello world" when they attempt something async).

When I look at the source in `/app/services/redux.js` I see that we are importing all 3 at the top and pass those in like so ...

```js
import ReduxService from 'ember-redux/services/redux';
import middlewares from '../middleware/index';

export default ReduxService.extend({ reducers, enhancers, middlewares });
```

Making me think it's logical that anyone extending this would then be required to import reducers/middlewares/enhancers or whatever they need to create the redux store manually.

TL;DR

I think this all makes sense and with a legit documentation section (under configuration) we could easily show how one might bolt on a custom middleware, or reducer, or enhancer.

I'll give @brettburley and @foxnewsnetwork time next week to think this over and provide feedback. I'm often in my own little world ;) so I'll step back and let them reflect on this (now) massive PR (sorry in advance guys)

toranb:
@bstro one last look at how we could show this in action. I took a step back and found this worked perfectly (as I import the reducers/enhancers myself and construct the saga middleware by hand)

```js
import redux from 'redux';
import ReduxService from 'ember-redux/services/redux';
import createSagaMiddleWare from 'redux-saga';
import root from '../sagas/index';

const { createStore, applyMiddleware, compose } = redux;

const sagaMiddleware = createSagaMiddleWare();

const makeStoreInstance = ({reducers, enhancers}) => {
  const middleware = applyMiddleware(sagaMiddleware);
  const createStoreWithMiddleware = compose(middleware, enhancers)(createStore);
  const store = createStoreWithMiddleware(reducers);
  sagaMiddleware.run(root);
  return store;
};

export default ReduxService.extend({
  reducers,
  enhancers,
  makeStoreInstance
});
```

foxnewsnetwork:
As far as I can tell, this LGTM; just be sure to at least bump minor (if not major) version so that we can always patch if users come up with unforeseen problems

bstro:
your usage example looks great to me. +1 -- thanks for your work on this! all your work has been super helpful for me and several people on my team.

addon/reducers/index.js
```diff
@@ -0,0 +1,19 @@
...
+// const foobar = (state, action) => {
+//   //reducer code here
+// };
+//
+// export default combineReducers({
+//   foobar
+// })
+import Ember from 'ember';
+
+export default function(combine) {
```
brettburley: I'm still a bit confused with what exactly this is exporting here, so I'm sure I'm missing something. Should this not just be a no-op reducer, e.g.:
```
export default function(state) {
  Ember.warn(...);
  return state;
});
```
otherwise, lgtm

toranb: Ah, I remember finding this in some redux source at one point. Because it's not actually intended for use we could default to the more simplistic bersuon you recommended.

When I'm back from vacation early next week I'll update this.

toranb:
@bstro @brettburley @foxnewsnetwork this has been released as part of ember-redux v2.10

bstro:
Nice!!

PR #126, opened 2017-05-04
Add "replaceReducer" method to redux service.

ssylvia:
This is the one method missing from the standard redux store: http://redux.js.org/docs/api/Store.html#store-methods. Needed if added reducers dynamically to the store (e.g. changing app state structure per route or dynamically loading component reducers).

### DIFF
--- a/app/services/redux.js
+++ b/app/services/redux.js
@@ -36,5 +36,8 @@ export default Ember.Service.extend({
   },
   subscribe(func) {
     return this.store.subscribe(func);
+  },
+  replaceReducer(nextReducer) {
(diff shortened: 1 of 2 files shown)


toranb:
@ssylivia this looks like something we can use to hot reload reducers that change also (still yet to expand the current hot loader I wrote beyond components but I think this method is what Dan used in the react ecosystem to inject reducers on the fly without a full page refresh if I remember right)

toranb:
@ssylvia this is now included in the v2.6.0 build of ember-redux! thanks again!

PR #127, opened 2017-05-08
[DEPENDENCY]: upgraded to ember-redux-shim/thunk-shim v2.0.0

toranb:
As part of the larger effort to make the entire `ember-redux` ecosystem of addons babel 6 friendly this bumps the 2 redux shims we use internally. This PR is up for feedback and should not be merged as it breaks node 4X users who are using `ember-cli` v2.4.3

I did some experimentation this morning to confirm node/ember-cli backwards compatibility and found that this PR would break anyone running on node v4.4.3

I have a full blown [example](https://github.com/toranb/olderreduxapp/commits/master) app up for those who have time to seek out alternative solutions but I believe this jump to babel 6 will indeed break someone not running node v5/v6

If we cannot find a workaround, this PR will become the first of many for the next major version of `ember-redux` -something I opened an [issue](https://github.com/ember-redux/ember-redux/issues/124) about last week


toranb:
Doing a little more investigation into supporting older ember-cli (mainly v2.4.3 in this case) I found node v5.12 works fine. I also noted that node v4.5 will compile (albeit with a ton of warnings/ and it was VERY slow) but ... when you boot the app using `ember s` you get a breaking error

```
ReferenceError: Unknown plugin "babel-plugin-transform-object-rest-spread" specified in "base" at 0, attempted to resolve relative to "."
```

Funny enough, after I ported another ember app up to `ember-cli-babel` 6.1.0 I learned that node 5 is actually not supported. So With this information I think the plan moving forward (maybe late June?) is to drop support for node 4/5 and focus on 6+

babel6

toranb:
Closing this out for now - I'll pick this back up in early July when the work for v3.0 kicks off

PR #129, opened 2017-06-01
Support factories for stateToComputed.

brettburley:
This allows reselect creators to be used, where a new instance
will be created for each instance of a connected component.

For more information on why this is useful, see: https://github.com/reactjs/reselect#sharing-selectors-with-props-across-multiple-components

This enables things like https://github.com/HeyImAlex/reselect-map to be used, which are critical for performance when mapping over large arrays in selectors (like a large list).

### DIFF
--- a/addon/connect.js
+++ b/addon/connect.js
@@ -60,6 +60,25 @@ function getAttrs(context) {
   return getProperties(context, keys);
 }
 
+/**
+  Return a wrapped stateToComputed function that can be used
+  regardless of wheter the input function is a static selector
+  or a factory.
+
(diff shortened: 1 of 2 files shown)


toranb:
Great work as always @brettburley !!! Really appreciate your involvement in this project !!!

toranb:
@brettburley I just added a quick shim for reselect-map so I can play with this end-to-end and throw up an example for people.

https://github.com/ember-redux/ember-reselect-map-shim

toranb:
This is now included in the v2.8.0 release

PR #133, opened 2017-07-03
Release v3.0.0

toranb:
This PR is the placeholder for a v3.0.0 release I'm planning later this month. On the agenda for breaking changes so far ...

1) I'm bumping the shims (breaking change upgrading from babel 5 => babel 6) 
2) Removing the handleChange and getAttrs methods (previously intimate apis in the 2x series)
3) officially supporting only ember-cli v2.4+ and node 6+
4) Removing the legacy connect export `ember-redux/components/connect`
5) Removing support for `export default {}` from the root reducer (combineReducers is required)

Anything else we should consider removing in the up coming release?


toranb:
Hmm maybe this would be a good time to work in the babel 6.6 stuff ...

```js
import { get } from '@ember/object';
import { computed } from '@ember/object';
```

What limitations would we have taking this on if any? Would this still work for ember-cli v2.4.3 users? Worst case we could take it for a spin :) just thought to ask if anyone is aware

toranb:
Sorry for the delay in landing this everyone - I'm still trying to work out the best solution for my monkey patched test cases :)

@bstro if this does land in the coming weeks is your team running node 6+ or did you run into a few blockers?

toranb:
I'm planning to publish the first beta early next week - if anyone has feedback on this before then let me know :)

PR #144, opened 2017-12-07
TESTS: added es2015 class based component to test drive connect

toranb:
Adding a simple test case to help drive the connect work ahead


toranb:
Added simple fork logic to support both ember object and es2015 class syntax

toranb:
@pzuraq I'm 99% ready to merge this PR and wanted to see if you could give it a brief look in the next 10/14 days or so (mostly to see if I stumbled upon anything of value / or if I've poorly implemented something)

toranb:
@brettburley I know it has been a while but I could sure use another set of eyes on one change in particular. As part of my testing initiative for es2015 class syntax I learned that the factory function support in ember-redux (from #129) wasn't working as I expected. I was able to get it working with a slight modification to the `getProps` function. It went from this (as it worked originally) ...

```js
  const getProps = wrapStateToComputed(() =>
    stateToComputed.call(this, redux.getState(), getAttrs(this))
  );
```

To this (tested /confirmed this does correctly update after stateToComputed)

```js
  const getProps = () => wrapStateToComputed(() => stateToComputed.call(this, redux.getState(), getAttrs(this)))();
```

brettburley:
Does the test case in `1acbf433859d9932e55ccfeabfe2633727db621f` cover the failure you're mentioning? I think the `state` parameter in the factory might be in the wrong place. I think it should be something like:
```
const makeUniqueSelector = () => {
  return (state) => state.low;
}

const stateToComputedFactory = () => {
  const selectItemHereOnly = makeUniqueSelector();
  // state is passed in to the generated function in subsequent calls.
  return (state) => {
    const low = selectItemHereOnly(state);
    return {
      low
    }
  }
}
```

I think the new implementation will actually invoke `stateToComputed` each time there is an update, re-generating a brand new function from the factory (i.e. calling `stateToComputedFactory` each update), but rather, we'd like the factory to be only called a single time for each component instance.

Does this make sense?

toranb:
@brettburley that was the approach I tried first but quickly I learned that state was `undefined` unless I moved it all the way up to `stateToComputedFactory`

screen shot 2018-01-06 at 7 41 23 am

brettburley:
@toranb so I think you're right; the factory handling is just completely broken... here's what I think should fix it: https://github.com/brettburley/ember-redux/commit/b6dfd9926073c7314a0f66018d90f27121c7d61b

Ultimately, if you used a factory, the returned function was never called with the updated state (in fact, it was called without any arguments). However, it should be called with the same arguments that a non-factory function would be called with.

toranb:
@brettburley thanks for confirming my suspicion and solving it! I've included your solution and it seems we are now inline w/ the mental model I have for "factory functions"

thank you 100x !!!

tests/dummy/app/components/clazz-basic/component.js
```diff
@@ -0,0 +1,19 @@
...
+});
+
+class MyClazz extends Component {
+  constructor() {
+    super(...arguments);
+    this.color = 'orange';
+  }
+}
+
+export default connect(stateToComputed, dispatchToActions)(MyClazz);
```
pzuraq: Long term this would probably be nice as a class decorator:

```js
@connect(stateToComputed, dispatchToActions)
export default class MyClazz extends Component {

}
```

But I think redux has a lot of experimenting it can do with decorators in general, and they still aren't 100% complete

addon/es2015-class.js
```diff
@@ -0,0 +1,27 @@
...
+
+  return class Connect extends WrappedComponent {
+
+    constructor() {
+      super(...arguments);
+      core.call(this, stateToComputed, dispatchToActions);
+    }
+
+    didUpdateAttrs() {
+      super.didUpdateAttrs(...arguments);
```
pzuraq: You probably will want to put this call to `didUpdateAttrs` in a guard, since the superclass may or may not have implemented `didUpdateAttrs`. We need to figure out what the solution is long term is Ember for this

PR #146, opened 2018-02-09
Add a link to the Reddit clone that is built in the intro video

benshine:
Let's make it super easy to find the Reddit clone that is built in the intro video.
 I haven't been able to verify this locally fix; how do I access this page when running locally?


toranb:
@benshine ah great idea! To run the website locally ...

cd into the `docs` directory
run the command `bundle install`
run the command `jekyll s`

note* this does require a ruby tool chain

Issue #180, opened 2019-03-01
State cannot be accessed in init hook in Ember 3.5.1


BobrImperator:
I use ember-redux on a project, and ocassionally I need to make a request from `init` hook which uses some data from the global state, currently I'm on `Ember 3.1.1` and trying to update to `3.5.1`, but the problem I've encountered holds me down.

Potentially changing a single line in `connect.js` fixes the problem:

`if (WrappedComponent.hasOwnProperty('extend')) {`  ->    `if (WrappedComponent.extend) {`
Because of when the call to the store is made. 
In `ember-object` [connector](https://github.com/ember-redux/ember-redux/blob/master/addon/ember-object.js), it happens before `WrappedComponent`'s `init` hook,
while for [`es-2015`](https://github.com/ember-redux/ember-redux/blob/master/addon/es2015-class.js), [`es-2019`](https://github.com/ember-redux/ember-redux/blob/master/addon/es2019-class.js) it happens later.

I've prepared a reproduction of the problem:
https://github.com/BobrImperator/redux-state-in-init

toranb:
@BobrImperator appreciate you opening this! I've spent a little time today trying to get this fixed without breaking older versions of ember and this seemingly trivial ask has proved to be a surprisingly uphill battle given all the churn es2015 class syntax has seen over the last 12+ months. I'll update you more next week when I take another look at this ;)

BobrImperator:
Actually, the `es-2019` version might work when the state call hook was to be moved above `super`, although I'm not sure if it might be breaking something.
The `es-2015` though seems to be failing because We cannot call the `state` call before `super`.

I'm also not sure, but I've read somewhere, that Ember will continue initializing with `init`, but that's an open topic right now.

toranb:
@BobrImperator I was able to solve it (even for 2.18 despite what Travis said during the PR build)

https://github.com/ember-redux/ember-redux/pull/182

I'll need to build a fresh 2.18 ember-cli app quick to confirm it's g2g but if so (fingers crossed) I'll publish a new version of this library next week. Can you confirm the master branch of this fixes your use case? I tried to take your example and throw that into the app as a test case (to confirm this is working for EmberObject from ember 2.18+)

BobrImperator:
I can confirm that this works for `3.5.1` `3.6` and latest `3.8.0` Ember!
Sweet, thank you very much!

[...]

PR #184, opened 2019-03-06
Move state calls up, add init for es-2015 class

BobrImperator:
@toranb Please take a look

Related to #180, as the issue still occurs.


BobrImperator:
Moving the calls for other life cycle hooks should be also considered, because if anyone were to access the state in DidReceiveAttrs for example, they'd end up with older state, if it happened to be updated.

toranb:
Just a heads up :) the es2015 init addition here is something I truly don’t want in the codebase (as es2019 solves that and was added for only that compat reason).

The root cause here is “why” doesn’t the conditional for EmberObject catch and connect up as it does wirh test/dev. Until we understand that more deeply I won’t be merging a bandaid over this to solve the immediate defect you ran into.

BobrImperator:
Sure, but I dunno If I'll be able to debunk the issue if  I know just as much.
The only things I know at the moment is that it hits the `es-2015` connector which has no init, and basically makes no call to the state at all.
Also apparently this occurs for Chrome and not for Firefox, and so it gets as much more interesting and the try/catch when calling the component returns
`Cannot read property 'proto' of undefined` instead of `Class constructor cannot be invoked without 'new'`

BobrImperator:
Also please take a look at the changes again, I couldn't help but think that `es-2015` file is redundant.

BobrImperator:
Alright! I DE-BUNKED it.
I've had a revelation, I simply ran tests in production environment on master branch in addon, and 1 of them failed, the one You've added for this issue. Then I ran tests on this PR branch, and it worked :)
I've modified ember-try to run in environment production right now.

toranb:
@BobrImperator  thanks for the extra effort on this today! I'm glad to see the nasty hacks I added last week gone and this works for ember 2.18+ so it should be g2g for everyone (for all envs dev/test/prod).

toranb:
@BobrImperator  I've published this on npm as v4.0.3

BobrImperator:
Lovely! Thank you very much!
<3

PR #191, opened 2019-08-11
Replace redux shims with ember-auto-import

jkusa:
This PR removes the need to use ember redux shims by replacing the dependencies with the direct npm packages and [ember-auto-import](https://github.com/ef4/ember-auto-import). This _should_ simplify the maintenance of the ember-redux project and allow users of ember-redux to work with direct npm packages.

A few notes:
- ember-auto-import has a dependency that requires >= node 8, therefore I bumped the node requirement in this repo
- I continued to make ember-redux-thunk-shim a default dependency controlled by the host app. Is this what we want going forward?. Let me know what you think.


toranb:
@jkusa excellent! I'm curious if we should release this as v5.0 or something despite it's backward compatibility but because I'm not writing ember myself on the regular I wanted your opinion on this :)

About thunk - I _think_ this is the right approach for now as some people might prefer saga or Rx over using thunk and I'd like that option to remain after this migration to `ember-auto-import`

note: I've also made you a full collaborator/admin so you can help make decisions like this in the future :) *assuming you want that responsibility

toranb:
@jkusa you might also rebase - recently testem had something break and I've just merged a fix for that

jkusa:
> I'm curious if we should release this as v5.0

That makes sense to me. This brings in ember-auto-import and the new node requirement.

Regarding thunk, I actually made a commit to make this transition more iterative. I kept the thunk shim in the blueprint so users would not have to use ember-auto-import in their host app. But IMHO, I would like default thunk middleware and dependency to be removed and allow users to pick their middleware without having to remove thunk from their package.json. Maybe a 5.0 release would be a good time to remove this.

 > note: I've also made you a full collaborator/admin so you can help make decisions like this in the future :) *assuming you want that responsibility

Awesome! I would love to help out.

toranb:
@jkusa suppose we make a splash w/ the v5.0 release as you suggest :) if you had to write a short blurb for the readme or docs about how to continue using thunk what would that look like?

I say we aim for exactly this^ and at the time of release we can give you all the glory ;) as I silently nodded my head while you did all the hard work

jkusa:
@toranb I rebased and added a commit to remove the default `redux-thunk` middleware.

> if you had to write a short blurb for the readme or docs about how to continue using thunk what would that look like?

I noticed you added an upgrade guide for 4.0 in the `releases` section of the repo. Do you want to do that again? I would say upgrade steps would look something like:

## Upgrading

ember-redux now uses ember-auto-import for redux dependencies. Node 8+ is required.

### Step 1 - Upgrade ember redux to v5.0.0

```
ember install ember-redux@5.0.0
```

### Step 2 - Remove the ember-redux-shim dependency

This ember shim is no longer necessary as ember-redux now uses ember-auto-import to bring in the redux library.

### Step 3 - If using redux-thunk middleware, explicitly configure it 

Previous versions of ember-redux configured redux-thunk by default. This now needs to be explicitly configured.

```js
//app/middleware/index.js

import thunk from 'redux-thunk';
export default [thunk];
```

### Step 4 - [Optional] Replace any ember-redux shims with the direct packages and ember-auto-import

 redux-thunk:
```bash 
$ ember install ember-auto-import
$ npm install --save-dev redux-thunk 
```

 redux-saga:
```bash
$ ember install ember-auto-import
$ npm install --save-dev redux-saga
```

toranb:
@jkusa this looks great! Usually the release process looks something like this

1) I could cut a branch (master) for release/4.x
2) merge this PR
3) append this to the changelog, and alter the package.json version to be 5.0
4) push that^ commit up
5) when it’s green ‘npm publish’

You at all interested in taking a run at this? If not, I should be open Sat morning this week

jkusa:
@toranb sure, I can cut the release. I should have some time over the weekend.

PR #195, opened 2019-12-09
DEPENDENCY: ember-cli-update to v3.14.3

toranb:
Upgrade -will need to validate the ember 3.4/3.5 compatibility before this is released

previous build for future triage  https://travis-ci.org/ember-redux/ember-redux/builds/622412422


toranb:
Decided to instead start a 6x by removing all the legacy ember object code we had (ie: ember apps running ember 3.5 and lower will not be supported). I don't think I'll release another major version bump with these changes until early next year but this should be acceptable at some point as ember continues to evolve and apps upgrade to the latest LTS

PR #198, opened 2020-02-03
Adding support for glimmer component

raghavbk:
To support Octane added the different code path for Glimmer Component. Most of the code taken from the spike oranb/octane-interop-redux@51e8fe1. Also added the factory function supported

I didn't find any better way to differentiates between glimmer and ember component 
`component.toString() === '@ember/component' && component.reopenClass;
`

Added required test to cover happy paths

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,9 @@ cache:
   yarn: true
 
 env:
+  - EMBER_TRY_SCENARIO=ember-3.6.X
+  - EMBER_TRY_SCENARIO=ember-3.8.X
+  - EMBER_TRY_SCENARIO=ember-3.12.X
   - EMBER_TRY_SCENARIO=ember-3.15.X
 
 before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,7 @@ cache:
   yarn: true
 
 env:
-  - EMBER_TRY_SCENARIO=ember-3.6.X
-  - EMBER_TRY_SCENARIO=ember-3.8.X
-  - EMBER_TRY_SCENARIO=ember-3.12.X
+  - EMBER_TRY_SCENARIO=ember-3.15.X
 
 before_install:
   - curl -o- -L https://yarnpkg.com/install.sh | bash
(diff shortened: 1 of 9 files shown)


toranb:
@raghavbk wow this was some amazing and speedy work !!! My only question is “could we keep ember try for 3.6+ or did those builds fail with this change? I wanted to release an update that would support only ember 3.? forward but I was hoping what we came up with wouldn’t break 3.6 if that makes any sense.

raghavbk:
It won't break with 3.6+ but build was failing as our test component is using glimmer

toranb:
@raghavbk ah that makes sense. Give this PR a look and see if you can wrap that glimmer test in something like this

https://github.com/ember-redux/ember-redux/pull/185/files

```js
    import { gte } from 'ember-compatibility-helpers';

    if (gte('3.15.0')) {
      //
    }
```

raghavbk:
@toranb Thank you, it worked :)

toranb:
Amazing work on this @raghavbk !!! I'll give this a spin tomorrow before a solid merge - you mind giving this branch a run on ember 3.6 and 3.12 along w/  octane at some point just to verify it's working with a mix of Ember (classic) and Octane (when available) components? Thanks so much for the great work on this !!!!

raghavbk:
Thanks @toranb, eger to use this in our project along with glimmer component

toranb:
@raghavbk sorry for the delay! I did some extensive testing with both ember v3.6 and v3.16.1 to confirm no obvious errors or regressions. I do think this is ready for prime time but because it's a fairly big shift I'm planning to release ember-redux v6.0 in a few days. I'll need to update a few things in the README and on the docs website to show that we now support both ember component AND glimmer component side by side ;)

Thanks for the amazing lift w/ this pull request dude !!!

toranb:
@raghavbk this should be up on npm now :) thanks again for the great work !!!

https://github.com/ember-redux/ember-redux/releases/tag/v6.0.0

raghavbk:
Thank you @toranb for quick turn around..

PR #209, opened 2021-02-01
Correctly import named exports from redux

austinezell:
Resolves issue: https://github.com/ember-redux/ember-redux/issues/208


toranb:
nice! From the looks of it we might need to bump to a 7.0 for people w/ this imported as it was - thoughts?

toranb:
any idea how we can restart that travis build? I don't see the usual options for a reload and that 1 broken build seems more like env flake than a real issue

